PR #2, opened 2016-06-10
Use absolute_import, unicode_literals

ericsuh:
Fixes a few type errors when working with unicode strings.


braincore:
Unfortunately doesn't look like it fixes the bug we're targeting, but we can still merge it. Can you add the future import to `stone_base.py` and `stone_serializers.py` as well.

ericsuh:
Hmm, seems to have had merging issues; I don't think I'm committing to the latest master. Resubmitted new pull request

PR #5, opened 2016-06-14
Check for name collisions between namespaces, types, routes and aliases.

scobbe:
This diff adds a check for name collisions between namespaces, types, routes and aliases in Stone specs. However, name clashes are not enforced across namespace, only within a given namespace.


scobbe:
Going to create pull request from forked repo

PR #6, opened 2016-06-14
Check for name collisions between namespaces, types, routes and aliases.

scobbe:
This diff adds a check for name collisions between namespaces, types, routes and aliases in Stone specs. However, name clashes are not enforced across namespace, only within a given namespace.


braincore:
Rebased after touch up in e0f22bb63d09bc4eaf7b321f3720966b5b0e3d2a. Thanks for doing this!

PR #7, opened 2016-06-16
Checks for variable names with the same name as Swift type keywords

scobbe:
In the event that a spec variable has the same name as a Swift type keyword, an underscore is appended to the variable name.


braincore:
Merged in with 3445469527f08bb352e83db1047195c3aa6ceaaf.

PR #11, opened 2016-07-27
Add some rudimentary PEP 484 (MyPy) annotations to the codebase.

wittekm:
Just wanted to clean up any errors it was throwing by default. 

(If you're curious what the mypy-run.sh script picks up, it is:
```
+ mypy --strict-optional --silent-imports --fast-parser setup.py stone/__init__.py stone/api.py stone/cli.py stone/cli_helpers.py stone/compiler.py stone/data_type.py stone/generator.py stone/lang/__init__.py stone/lang/exception.py stone/lang/lexer.py stone/lang/parser.py stone/lang/tower.py stone/target/__init__.py stone/target/helpers.py stone/target/js_client.py stone/target/js_helpers.py stone/target/python_client.py stone/target/python_helpers.py stone/target/python_rsrc/__init__.py stone/target/python_rsrc/stone_base.py stone/target/python_rsrc/stone_serializers.py stone/target/python_rsrc/stone_validators.py stone/target/python_types.py stone/target/swift.py stone/target/swift_client.py stone/target/swift_helpers.py stone/target/swift_types.py test/test_cli.py test/test_generator.py test/test_python_gen.py test/test_stone.py test/test_stone_internal.py
```

### DIFF
--- a/stone/lang/lexer.py
+++ b/stone/lang/lexer.py
@@ -5,6 +5,8 @@ import os
 
 import ply.lex as lex
 
+from typing import Tuple
+
 class MultiToken(object):
     """Object used to monkeypatch ply.lex so that we can return multiple
     tokens from one lex operation."""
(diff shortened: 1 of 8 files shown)


stone/lang/lexer.py
```diff
@@ -100,41 +102,31 @@ def test(self, data):
         'KEYWORD',
         'PATH',
         'DOT',
-    )
 
-    # Whitespace tokens
-    tokens += (
+        # Whitespace tokens
```
braincore: Did you have to move these comments within the tuple for mypy?

wittekm: I did not. It had trouble automatically figuring out that the initial `tokens` tuple was a Tuple[many strings], so I decided to jsut collapse it all together.

There's two solutions:
* add `# type: Tuple[str, ...]` (which i did elsewhere in this code review)
* just collapse them all together

happy to go back to the original approach :)

braincore: I'm more just wondering how smart mypy is. Could you just have typed the first block as `# type: Tuple[str, ...]` and left all concatenations untouched?

braincore:
Did you install `mypy` in PY2 or PY3?

wittekm:
mypy is strictly a Python 3 application, though you can use it to typecheck both python 2 and python 3 code (based on whether the `--py2` flag is turned on).

Since Stone works for both, I can change `mypy-run.sh` to run it with the --py2 flag on and off.

stone/lang/lexer.py
wittekm: Took care of that. The smaller the diff, the better :)

PR #12, opened 2016-08-01
Fully type-annotate generator.py and api.py

wittekm:
I'm unsure about ApiNamespace's `self.data_types` being a list of UserDefined, but that seems to be the greatest common divisor that has a `data_type.namespace` as referenced on line 263.

(meta-discussion: no, I haven't gotten around to actually generating type-stubs yet, unfortunately)

### DIFF
--- a/stone/api.py
+++ b/stone/api.py
@@ -1,28 +1,44 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
+from typing import (
+    Dict,
+    List,
+    Optional,
+    Union,
+    cast,
+)
 
 from collections import OrderedDict
 from distutils.version import StrictVersion
 import six
 
 from stone.data_type import (
+    Alias,
+    List as DataTypeList,
+    Nullable,
+    Struct,
+    UserDefined,
     doc_unwrap,
     is_alias,
     is_composite_type,
     is_list_type,
     is_nullable_type,
 )
+from stone.lang.parser import StoneRouteDef
 
+NamespaceDict = Dict[str, 'ApiNamespace']
 
 class Api(object):
     """
     A full description of an API's namespaces, data types, and routes.
     """
     def __init__(self, version):
+        # type: (str) -> None
         self.version = StrictVersion(version)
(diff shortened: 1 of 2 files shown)


stone/api.py
```diff
@@ -1,28 +1,42 @@
 from __future__ import absolute_import, division, print_function, unicode_literals
+from typing import (
+    Dict,
+    Optional,
```
braincore: Alphabetize.

braincore: Alphabetize (`Optional` under `List`)

stone/api.py
```diff
@@ -181,24 +207,27 @@ def add_alias(alias):
...
         assuming it's a user-defined type.
         """
-        data_types = set()
+        data_types = set()  # type: Set[UserDefined]
         for route in self.routes:
             for dtype in (route.arg_data_type, route.result_data_type,
                           route.error_data_type):
                 while is_list_type(dtype) or is_nullable_type(dtype):
-                    dtype = dtype.data_type
+                    dtype_as_list = cast(DataTypeList, dtype)
```
braincore: In this case, it could be a `List` or `Nullable` object. Can you cast to a Union of the two?

wittekm: Sure. 

As a side note, this was done mainly because the base class `DataType` doesn't have a `.data_type` property; perhaps this implies that we should have some typecheckable way to say "this has a concrete .data_type"

braincore: I think it should be considered more of a coincidence that the two have a `data_type` property. I don't, for example, think that `List` and `Nullable` should implement a hypothetical interface (not that you can in Python) called "wraps another type". So if anything, I'd be more comfortable with writing the while-loop body so that it explicitly checks which type it is (`List` or `Nullable`)and accesses the wrapped type accordingly.

I actually think it's worse that they

braincore:
Looks good! Feel free to rebase it in. Thanks a ton for doing this.

braincore:
(Also, please squash your commits)

PR #13, opened 2016-08-21
Better JSDoc Types: Emit @typedefs, use standard primitive types, & more

jvilk:
This PR changes the following in the Javascript generator, and addresses [this issue in the Dropbox JS SDK repository](https://github.com/dropbox/dropbox-sdk-js/issues/50):

## Changes

Changes to `js_client` generator:

* Changes primitives to use lowercase type names, as in the [JSDoc documentation](http://usejsdoc.org/tags-type.html).
  * Conveniently, TypeScript uses the same lowercase type names for primitives.
* Changes `Void` type to `void` rather than `null`.
  * While JSDoc doesn't have an official void type, [ESLint's valid-jsdoc linting rule m the `void` type](http://eslint.org/docs/rules/valid-jsdoc). It also lines up with TypeScript's `void` type.
* Changes routes to return a `Promise` object, with success and error types (e.g. `Promise.<FilesMetadata, Error.<FilesAlphaGetMetadataError>>`)
* List types are now emitted as `Array.<ComposedType>` rather than just `Array`.
* Timestamps are emitted as `Timestamp`, which it JSDoc `@typedef`s to `string`
* Moved `-e` parameter to the new `js_types` generator...

Adds a `js_types` generator that:

* Generates [JSDoc typedefs](http://usejsdoc.org/tags-typedef.html) for referenced struct and union types.
* Nullable fields on typedefs are now emitted as "optional" fields.
* Takes in `--extra-args` to add extra parameters to route argument types.

## Possible Improvements

* Emit `@typedef` for alias types. For now, we unwrap all aliases.
* Include stone attributes in field / type descriptions, e.g. regular expressions on string types.

## Example Output

[Here is the Dropbox Javascript SDK documentation](https://github.com/dropbox/stone/files/457956/dropbox-sdk-js-docs.zip) and [here is the SDK route/typedef code](https://github.com/dropbox/stone/files/457955/dropbox-js-sdk-generated-code.zip) generated with these generator changes.

Since this is a rather large change to your generator, I'm expecting you to have plenty of comments and nits regarding the code I've written. Python isn't my primary language, so it's possible that some of my code isn't idiomatic / doesn't follow your style guide.

Let me know what you think!


jvilk:
RE:Union types, It looks like the Java SDK, which currently doesn't have an open source generator, has deserialization logic that treats "collapsed" and "non-collapsed" unions differently.

The HTTP documentation shows all error types, which are all union types AFAIK, as having the following structure:

```json
{
  "error_summary": "string",
  "error": {
    ".tag": "union tag",
    // extra properties on this particular tag, is any
  }
}
```

But `error_summary` isn't a field mentioned in the Stone data types, and `.tag` isn't documented elsewhere for union types. So I'm still not completely sure how to treat them when generating JSDoc.

braincore:
Elated to see this, but it will take me some time to review this week.

To answer your last question, we should have a type definition like this:
```
struct Error<T>  # T is the error type for the route
    error_summary String
    error T
    user_message UserMessage

struct UserMessage
    text String
    locale String
```

However, we don't because stone doesn't have support for generics. To get around this, these types are simply handwritten into our SDKs and thus absent from our specs.

braincore:
Re union types: Specifying a union variant with an object that has a `.tag` key will always work. The string version that you've seen is a "compact form" that only works for variants of void type.

We introduced the compact form to make it easier to specify when using a command-line tool like curl or httpie. We also did it for aesthetic reasons, which is why you'll see our documentation page use it for example arguments, but not for results; we always return the verbose form.

There's a [short blurb here in stone's documentation](https://github.com/dropbox/stone/blob/master/doc/json_serializer.rst#L180).

braincore:
Is it JSDoc or TypeScript that isn't happy with having multiple typedef files?

stone/target/js_client.py
```diff
@@ -4,6 +4,8 @@
 import json
 import six
 import sys
+import os
```
braincore: Alphabetize imports.

stone/target/js_client.py
```diff
@@ -24,6 +26,13 @@
           'contains all of the routes.'),
 )
 _cmdline_parser.add_argument(
+    'typedef_filename',
```
braincore: Are you sure we should require this argument rather than generating additional type information only if a filename is specified?

stone/target/js_client.py
```diff
@@ -73,8 +85,33 @@ def generate(self, api):
...
+                existing_typedefs = typedef_file.read()
+            # Looking for patterns like so:
+            # @typedef {base type} TypeName
+            for m in re.finditer('@typedef\s*\{[\w+]+\}\s*(\w+)\s*\n', existing_typedefs):
+                skip_typedefs.add(m.group(1))
+
+        with self.output_to_relative_path(self.args.typedef_filename):
+            # Re-emit existing typedefs, ensuring that a newline occurs
+            # at the end.
+            if existing_typedefs[-1:] is not '\n':
```
braincore: `[-1:]` asks for all characters after and including the last character. Since we know that's only one character, you should just use `[-1]`.

You should use `!=` instead of `is not` which is normally reserved for singletons (`None`). It's less preferred, but still reasonable here to have used `existing_typedefs.endswith('\n')`.

stone/target/js_client.py
braincore: Also, perhaps we should use the same pattern that's used for python and swift. That is, we have two generators: `python_types` and `python_client`. The former is invoked once to generate all types (in this case, we would just generate the typedefs). The latter is invoked twice (once for user routes, once for team routes).

stone/target/js_helpers.py
```diff
@@ -45,8 +48,53 @@ def fmt_obj(o):
...
 
 
-def fmt_type(data_type):
-    return _type_table.get(data_type.__class__, 'Object')
+def _create_user_defined_type(data_type, user_defined_types, process_doc, extra_parameters):
+    """
+    Generates a JSDoc @typedef for a user generated type (struct or union type).
+    """
+    type_docstring = '/**'
+    type_docstring += '\n * %s' % data_type.doc if data_type.doc else ''
```
braincore: You should use the `emit_wrapped_text` helper to bound line length. I think we've standardized on 80 characters. @RileyTomasek to confirm.

RileyTomasek: yeah, 80 characters should be the maximum line length.

jvilk:
> Is it JSDoc or TypeScript that isn't happy with having multiple typedef files?

Multiple typedef files are fine, but we need a mechanism to avoid re-emitting the same type twice in different files. In the case of Dropbox, there are certain structs that are shared amongst the team and non-team routes. Simply emitting every typedef referenced by the routes would result in typedef duplicates.

> Also, perhaps we should use the same pattern that's used for python and swift. That is, we have two generators: python_types and python_client. The former is invoked once to generate all types (in this case, we would just generate the typedefs). The latter is invoked twice (once for user routes, once for team routes).

That... actually sounds like a great idea! I'll do that.

stone/target/js_helpers.py
jvilk: Roger. Now that we've decided to have a separate generator for types, I can use generator methods to properly emit typedefs.

jvilk:
@braincore / @RileyTomasek I have updated the PR with a new `js_types` generator. The code is much cleaner now, and union types are properly emitted.

I updated the PR description, and linked to new documentation / JS files generated with these changes.

Let me know what you think!

doc/builtin_generators.rst
```diff
@@ -29,21 +29,21 @@ command-line interface (CLI)::
     usage: stone [-h] [-v] [--clean-build] [-f FILTER_BY_ROUTE_ATTR]
                  [-w WHITELIST_NAMESPACE_ROUTES | -b BLACKLIST_NAMESPACE_ROUTES]
                  generator output [spec [spec ...]]
-
```
jvilk: I configured Visual Studio Code to auto-remove extraneous whitespace at EOL. Sorry for cluttering up the PR! lmk if I should reintroduce this whitespace.

stone/target/js_types.py
```diff
@@ -0,0 +1,213 @@
...
+
+_header = """\
+// Auto-generated by Stone, do not modify.
+/**
+ * An Error object returned from a route.
+ * @typedef {Object} Error
+ * @property {string} error_summary - Text summary of the error.
+ * @property {T} error - The error object.
+ * @property {UserMessage} user_message - User-friendly error message.
+ * @template T
```
jvilk: JSDoc itself doesn't seem to recognize `@template`, but it's a JSDoc tag used by Google Closure Compiler to denote generics and keeping it around doesn't hurt doc generation. Googling suggests that some IDEs, like WebStorm, recognize this tag.

jvilk:
@braincore `js_types` is not emitting Dropbox's `--extra-args` data properly. While debugging, I notice that `route.attrs` is *empty* on all routes, preventing `_parse_extra_args` from working properly.

Do I need to toggle some type of generator configuration option to enable `route.attrs`?

jvilk:
I found a few bugs, and updated the documentation / generated code links in the original post.

braincore:
@jvilk: You'll need to use the `-a` argument to stone. Only then are the specified route attributes available to generators.

jvilk:
@braincore Ahh, I overlooked that before. OK, I verified that the generator properly processes extra arguments and adds them to the relevant data type(s).

jvilk:
*poke* I'm assuming everyone is busy. Let me know if there's anything else I should do for this PR (besides squash the commits once reviewed).

stone/target/js_types.py
```diff
@@ -0,0 +1,211 @@
...
+ * @property {string} locale
+ */
+
+ /**
+  * @typedef {string} Timestamp
+  */
+"""
+
+
+class JavascriptTypeGenerator(CodeGenerator):
```
braincore: nit: Type -> Types

stone/target/js_types.py
```diff
@@ -0,0 +1,211 @@
...
+                            continue
+                        attr_val = route.attrs[attr_key]
+                        if attr_val in extra_args[attr_key]:
+                            extra_parameters.append(extra_args[attr_key][attr_val])
+                if len(extra_parameters) > 0:
+                    extra_args_for_types[route.arg_data_type] = extra_parameters
+
+        return extra_args_for_types
+
+    def _generate_type(self, data_type, extra_parameters = []):
```
braincore: 1. No spaces around the `=`.
2. Standard practice is to avoid using a list as a default argument. You can read why here: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

stone/target/js_types.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        union_name = fmt_type(union_type)
+        self._emit_jsdoc_header(union_type.doc)
+        self.emit(' * @typedef {Object} %s' % union_name)
+        variant_types = []
+        for variant in union_type.all_fields:
+            variant_types.append("'%s'" % variant.name)
+            variant_data_type, _, _ = unwrap(variant.data_type)
+            # Don't emit fields for void types.
+            if is_void_type(variant_data_type) == False:
```
braincore: More idiomatic to write `if not is_void_type(variant_data_type):`

stone/target/js_types.py
```diff
@@ -0,0 +1,211 @@
...
+                variant_doc = ' - Available if .tag is %s.' % variant.name
+                if variant.doc:
+                    variant_doc += ' ' + variant.doc
+                self.emit_wrapped_text('@property {%s} [%s]%s' % (fmt_type(variant_data_type), variant.name, variant_doc), prefix = ' * ')
+        jsdoc_tag_union = '(' + '|'.join(variant_types) + ')' if len(variant_types) > 1 else variant_types[0]
```
braincore: Does typescript properly interpret this as an enum?

braincore:
Sorry for the delay. I just hooked this up into my local `dropbox-sdk-js` and re-generated the jsdocs, and it's a great improvement.

The one omission I see is [struct polymorphism](https://github.com/dropbox/stone/blob/master/doc/lang_ref.rst#struct-polymorphism). For example, the output of `filesGetMetadata` is listed as `FilesMetadata`. But really, the output will be either `FilesFileMetadata`, `FilesFolderMetadata`, or `FilesDeletedMetadata` (all of which in the stone spec share `Metadata` as a base class). This isn't conveyed in the documentation and can be confusing.

braincore:
Once you address struct polymorphism, and the minor n its, I think we'll be good to go! Thanks again!

stone/target/js_types.py
jvilk: Yup, but TypeScript doesn't actually take JSDoc type annotations into consideration. Previously, I standardized the primitive types to lowercase because it was consistent with both JSDoc documentation and TypeScript. I experimented with a tool that converts JSDoc into a TypeScript definition file (which specifies TypeScript typings for non-TypeScript libraries), but the output was not up to par with what could be achieved with a stone TypeScript definitions generator.

TypeScript will actually support tagged unions in its upcoming 2.0 release (there's a release candidate out now). With this addition, all of Stone's data types map cleanly onto TypeScript's type system! I [wrote a proper TypeScript definition file generator for stone](https://github.com/jvilk/stone/blob/tsd-types-generator/stone/target/tsd_types.py) that generates these tagged unions, but they currently [expose a small bug in TypeScript that will be fixed prior to a proper 2.0 release](https://github.com/Microsoft/TypeScript/issues/10530). Once working, TypeScript will 100% properly tag the arms of a `switch` statement over the `.tag` field, and can determine statically when the switch does not cover all enum variants.

I am happy to open up a PR with the TypeScript typings generator, btw. [I documented what it will cost from a maintenance standpoint to ship TypeScript typings with the Dropbox JS SDK in this issue](https://github.com/dropbox/dropbox-sdk-js/issues/65). If there's no interest from Dropbox, I'll submit the generated types to DefinitelyTyped once the TypeScript bug is fixed, so the people will get the goods either way. :)

stone/target/js_types.py
jvilk: Ah! I've actually implemented a Python interpreter, and forgot about this subtlety. This optional argument is never actually used, so I can remove it.

jvilk:
> The one omission I see is struct polymorphism.

If I understand correctly: I'll have to check if any of a type's parent types is polymorphic. If it is, the type has an implicit `.tag` field. Then, when emitting routes, I'll have to check if the type is polymorphic; if it is, I emit an OR of all possible subtypes.

jvilk:
Actually, this is a tad confusing. You're saying that `FilesFileMetadata` will look like the following?

```json
{
  ".tag": "file",
  "name": ...,
  "id": ...,
   (etc)
}
```

In a standard union, `.tag` points to a field name that contains the union variant's data. In this case, `.tag` doesn't point to any field, but instead dictates which subtype it is.

Is this interpretation correct?

braincore:
Yes, that's correct. The (short) serialization docs for them are here: https://github.com/dropbox/stone/blob/master/doc/json_serializer.rst#L71

How well will this map to JSDocs?

jvilk:
@braincore Well, there's two ways to handle this.

1. Make the parent type have all of the child type's fields as optional fields. The docs for each field will mention which `.tag` makes it appear.

2. Make each child type have a `.tag` field. Since they are the child of one of these polymorphic structs, they always have a `.tag` field set to some value (or values if they are the child of multiple polymorphic structs, I guess). Make the return value of *closed* unions be the union of all of the subtypes. Make the return value of *open* unions be the union of all of the subtypes with the parent type.

I think the second option is better.

jvilk:
@braincore OK, I've addressed the nits, and added support for structural polymorphism! The links to generated docs / JavaScript have been updated in the OP.

jvilk:
@braincore btw, with this change, structs with enumerated subtypes that are *closed* (which I believe is the criteria for `is_catch_all` to be false?) are never directly referenced in the output JSDoc, since any reference to them (via field, argument, error type...) is replaced with a union of all of its subtypes.

In this case, should I not emit a typing for the struct?

stone/target/js_types.py
```diff
@@ -159,7 +161,35 @@ def _generate_struct(self, struct_type, extra_parameters = [], nameOverride = No
...
+                # This struct is the parent to multiple subtypes. Determine all of the possible
+                # values of the .tag property.
+                tag_values = []
+                for tags, _ in struct_type.get_all_subtypes_with_tags():
+                    for tag in tags:
+                        tag_values.append('"%s"' % tag)
+
+                jsdoc_tag_union = fmt_jsdoc_union(tag_values)
+                self.emit_wrapped_text('@property {%s} .tag - Tag identifying the subtype variant.' % jsdoc_tag_union)
+            else:
```
braincore: This case is unnecessary. If one of the subtypes is directly referenced, then there will be no `.tag` attribute since it isn't ambiguous. An example of this is the `filesUpload` route. It will return a `FilesFileMetadata` object (you can only upload a file, not a folder), and there's no `.tag` attribute.

Also, we don't support multiple layers of inheritance with enumerated subtypes. It's always only a single root struct and one layer of children (we might expand the design later).

braincore:
Verified that the enumerated subtypes output looks correct in JSDocs! Last bit is to remove some unnecessary logic, and then I think we're ready to merge this!

stone/target/js_types.py
jvilk: > It's always only a single root struct and one layer of children

OK; some of the helper functions in the code seemed to assume otherwise, so I went with a general design.

> If one of the subtypes is directly referenced, then there will be no .tag attribute since it isn't ambiguous.

Oh, that complicates things. I could do the following:

1. Make `.tag` an optional field on these data types.
2. Make a second data type for `FilesFileMetadataVariant` or something that has `.tag`, while `FilesFileMetadata` does not.
3. Ignore the `.tag` option entirely.

I think 1 is the cleanest solution. I can put a comment, like "Tag identifying this subtype variant. This field is only present when needed to discriminate between multiple possible types."

jvilk: Also, unless you have a good reason for me not to, I think I'll keep the more general code in there as it seems consistent with other design choices. For example, `is_member_of_enumerated_subtypes_tree` recursively climbs up the inheritance hierarchy.

jvilk:
OK, I think I came up with a useful compromise. I don't think the logic is unnecessary. When the docs points to `FilesFileMetadata | FilesFolderMetadata` as a route's return value, a developer needs to know how to discriminate between the two. Omitting the `.tag` field on those types entirely would be confusing in those cases.

As a compromise, I made it an optional field with a clarifying description. Does this work for you, or do you have a better idea in mind?

stone/target/js_types.py
braincore: Okay, that's fine :)

braincore:
Haven't had a chance to test this today. But in preparation for landing, please squash your commits and write a descriptive commit log.

braincore:
Just tested. Looks great! Just waiting for the squash.

jvilk:
Squashed with descriptive commit message! (Which could probably be spun off into a doc if you feel like maintaining the description.)

braincore:
Thanks for the detailed commit message! Rebased in. I'll be updating the Dropbox JS SDK soon.

PR #15, opened 2016-09-20
Implement equality and hashing for unions

icedtrees:
Unions should be considered equal to each other if they subclass each other and have the same tag and value. See proposal at https://paper.dropbox.com/doc/APIv2-Union-and-namedtuple-2MZPn0EpJVfYKw5pa4boo?t=549589594649114#:t=549589594649114

### DIFF
--- a/stone/target/python_rsrc/stone_base.py
+++ b/stone/target/python_rsrc/stone_base.py
@@ -34,6 +34,20 @@ class Union(object):
         self._tag = tag
         self._value = value
 
+    def __eq__(self, other):
+        # Also need to check if one class is a subclass of another. If one union extends another,
+        # the common fields should be able to be compared to each other.
+        return (
+            isinstance(other, Union) and
+            (isinstance(self, other.__class__) or isinstance(other, self.__class__)) and
+            self._tag == other._tag and self._value == other._value
+        )
+
+    def __ne__(self, other):
+        return not self == other
+
+    def __hash__(self):
+        return hash((self._tag, self._value))
 
 class Route(object):
 
--- a/test/test_python_gen.py
+++ b/test/test_python_gen.py
@@ -550,6 +550,19 @@ union_closed U
     t0
         "Sample field doc."
     t1 String
+    t2
+
+union UOpen extends U
+    t3
+
+union_closed UExtend extends U
+    t3
+
+union_closed UExtend2 extends U
+    t3
+
+union_closed UExtendExtend extends UExtend
+    t4
 
 union V
     t0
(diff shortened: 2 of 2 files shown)


stone/target/python_rsrc/stone_base.py
```diff
@@ -34,6 +34,16 @@ def __init__(self, tag, value=None):
         self._tag = tag
         self._value = value
 
+    def __eq__(self, other):
+        # Also need to check if one class is a subclass of another. If one union extends another,
+        # the common fields should be able to be compared to each other.
+        return (
+            (isinstance(self, other.__class__) or isinstance(other, self.__class__)) and
```
cakoose: Why not just check that the classes are the same?  Do we ever want a value to be "equal" to a value of a subclass?

icedtrees: Yes. See discussion at https://tails.corp.dropbox.com/D227666:

union A
   t0

union B extends A
   t1

A(t0) should be equal to B(t0).

cakoose: Oh, that makes sense.  Though now I'm a bit worried about "identical" values having different underlying classes.

When someone instantiates says `B.t0`, maybe we should return `A.t0`?

stone/target/python_rsrc/stone_base.py
```diff
@@ -34,6 +34,16 @@ def __init__(self, tag, value=None):
...
+            self._tag == other._tag and self._value == other._value
+        )
+
+    def __hash__(self):
+        return hash((self._tag, self._value))
```
cakoose: The Python docs say to implement `__ne__` if you implement `__eq__`.  (It's fine to implement `__ne__` by calling `__eq__`.)

stone/target/python_rsrc/stone_base.py
```diff
@@ -34,6 +34,16 @@ def __init__(self, tag, value=None):
         self._tag = tag
         self._value = value
 
+    def __eq__(self, other):
```
cakoose: If someone passes in an instance of `object()`, this function will fail with a runtime error, right?  I'm not sure what the contract of `__eq__` is, but maybe you're required to return `false` in that case?

stone/target/python_rsrc/stone_base.py
braincore: In the simplest case of referencing the auto-populated `B.t0`, it is already a reference to `A.t0`. But, it's possible to manually instantiate `B` to `t0`, and it would take some generator work to reject that. I think out of scope for this diff. Also, the deserializer uses the manual instantiation method :( so that'll have to change.

stone/target/python_rsrc/stone_base.py
braincore: You should return false if `other` isn't a subclass of `Union`.

braincore:
Sorry for the delay, was on PTO for a couple of weeks. Once you address @cakoose's comments, this is good to go.

icedtrees:
Sorry for late follow-up! Addressed comments. Thanks for thorough review!

I'm not part of the Dropbox github team, so you'll have to add me or merge the PR.

braincore:
Can you please squash your commits? I'm going to try this new "Rebase and Merge" GitHub feature :)

icedtrees:
Done!

PR #16, opened 2016-10-05
Generate PEP 484 .pyi type stubs (ready for review)

wittekm:
For review, I'd recommend the following:
- Look at the test first to get an idea of how the output looks.
- Compare the code flow of the new generator with the layout of the existing `python_types.py` - very similar.

Definitely needs more testing before we flip it on at Dbx, but it's a good start.


wittekm:
Interesting - I added `'mock>=2.0.0', ` to install_reqs, but the py27 env still isn't picking it up.

wittekm:
Hypothesis: "if I add `mock` to `tox.ini`'s deps section, we should be good
Conclusion: Nope, same problem

wittekm:
(ping)

stone/target/python_type_mapping.py
```diff
@@ -0,0 +1,88 @@
+from typing import (
+    Any,
+    Callable,
+    Dict,
+)
+
+if False:  # Versioning issues :(
```
braincore: What's the deal here?

stone/target/python_type_mapping.py
```diff
@@ -0,0 +1,88 @@
...
+
+# Unfortunately these are codependent, so I'll weakly type the Dict in Callback
+Callback = Callable[[ApiNamespace, DataType, Dict[Any, Any]], str]
+OverrideDefaultTypesDict = Dict[DataTypeCls, Callback]
+
+"""
+NOTE: Due to some trickiness with MyPy, we sometimes choose to do a pure `isinstance` check instead
+of the much nicer looking `is_something_type` check.
+
+Example: in this Alias check, if we used `is_alias_type, you'd get the following message:
```
braincore: Ack, so unfortunate. The asymmetry is jarring, though it does make sense. Can we continue to use the `is_*` versions but add a cast (https://www.python.org/dev/peps/pep-0484/#casts) in the `elif` block?

Maybe in the future we'll switch entirely to using explicit `isinstance` but that'll be a larger refactor.

setup.py
```diff
@@ -18,6 +18,7 @@
     'ply>=3.4',
     'six>=1.3.0',
     'typing>=3.5.2',
+    'mock>=2.0.0',
```
braincore: Alphabetize please.

stone/target/python_type_mapping.py
```diff
@@ -0,0 +1,88 @@
+from typing import (
```
braincore: Do you think it would be reasonable to say that all typing imports should be prefixed with `T`?

For example,
```
from typing import (
    TAny,
    TCallable,
    TDict,
)
```

Standardizing this will help reduce the name conflicts caused by typing a typing system :)

test/test_python_type_stubs.py
```diff
@@ -0,0 +1,161 @@
...
+    Boolean,
+    List as StoneList,
+    UInt64,
+    Union,
+    UnionField,
+    Void,
+)
+from stone.target.python_type_stubs import PythonTypeStubsGenerator
+
+class TestPythonTypeStubsHelpers(object):
```
braincore: This class requires no state, so better to make this a collection of top-level functions.

test/test_python_type_stubs.py
```diff
@@ -0,0 +1,161 @@
...
+        ns.add_data_type(u1)
+
+        return ns
+
+    def _api(self):
+        api = Api(version="1.0")
+        return api
+
+class TestPythonTypeStubs(unittest.TestCase, TestPythonTypeStubsHelpers):
+    def test__generate_base_namespace_module__generates_stuff(self):
```
braincore: Why the double underscores in the name?

stone/target/python_types.py
```diff
@@ -196,6 +196,8 @@ def _docf(self, tag, val):
     def _python_type_mapping(self, ns, data_type):
         """Map Stone data types to their most natural equivalent in Python
         for documentation purposes."""
+        return map_stone_type_to_python_type(ns, data_type)
+        # TODO delete
```
braincore: Don't forget your TODO.

stone/target/python_type_mapping.py
wittekm: Alternatively I could just do `import typing` and just refer to them as `typing.Callable`

wittekm:
Implemented/responded to all of @braincore 's comments except the "all typing imports should be prefixed with T?"
(Will squash commits once I get a shipit.)

test/test_python_type_stubs.py
wittekm: Personal preference - puts a separator between <what's being tested> and <what's the expected behavior>, makes it easier to parse as a human IMO

(the equivalent of something like an RSpec like
```
describe "generate_base_namespace_module" do
  it "generates stuff" do
    test goes here
  end
end
```
)

posita:
> … all typing imports should be prefixed with T?

Whoops. 😳 I guess I need to start doing this too.

@wittekm/@braincore, is this ready to go? I can rebase and merge if it's ready.

wittekm:
It is ready to go! I can send you some examples (outside GitHub) of what
the results look like when run on the Dropbox rServer repo.

braincore:
Ping to resolve conflicts.

test/test_python_type_stubs.py
```diff
@@ -0,0 +1,332 @@
...
+    ns.add_data_type(struct)
+
+    return ns
+
+def _make_namespace_with_a_union():
+    # type: (...) -> ApiNamespace
+    ns = ApiNamespace('ns_with_a_union')
+
+    u1 = Union(name='U1', namespace=ns, token=None, closed=True)
+    u1.set_attributes(
```
braincore: We need a union that has a variant with type other than void. Non-void variants will generate an additional `get_variant()` function along with the `is_variant()` that you already type.

wittekm:
Alright - got all the Typing-related conflicts out of the way, and added non-void Union support like
```
            class Shape(bb.Union):
                point = ...  # type: Shape

                def is_point(self) -> bool: ...

                def is_circle(self) -> bool: ...

                @classmethod
                def circle(cls, val: float) -> Shape: ...

                def get_circle(self) -> float: ...
```

setup.py
```diff
@@ -15,6 +15,7 @@
 from setuptools import setup
 
 install_reqs = [
+    'mock>=2.0.0',
```
braincore: While I've historically put `mock` under `install_reqs`, these days we can just make it a test dependency right? @posita has done a good job of making that separation clear.

stone/target/python_type_mapping.py
```diff
@@ -0,0 +1,91 @@
...
+    is_nullable_type,
+    is_string_type,
+    is_timestamp_type,
+    is_user_defined_type,
+    is_void_type,
+)
+from stone.target.python_helpers import class_name_for_data_type
+from stone.typing_hacks import cast
+
+MYPY = False
```
braincore: Gosh, I'm getting used to the niceness of doing everything new in Python 3, that I think I might just make Stone's generator-interface PY3 only (of course we still need to generate PY 2 code...).

braincore:
I'm pretty happy with this! What's the goal in terms of checking the output into the Dropbox server repo? Are you going to create a server repo diff that updates stone to include this generator and fix the identified problems?

Last thing, could you add a blurb about the generator here: `https://github.com/dropbox/stone/blob/master/doc/builtin_generators.rst`

We haven't done a good job keeping that reference updated (only one of the python gens is in there), but now is as good a time as any to start!

setup.py
posita: Well, making it _clearer_ anyway…. :smirk: Consider rebasing on top of #41 (if it isn't yet merged) and adding `mock>=2.0.0` to `test/requirements.txt` instead. That will ensure it's only installed when performing tests (either via `setup.py test` or via `tox`).

My apologies for not making this approach available earlier.

stone/api.py
```diff
@@ -96,7 +96,7 @@ class ApiNamespace(object):
     """
 
     def __init__(self, name):
-        # type: (str) -> None
+        # type: (typing.Text) -> None
```
posita: @braincore, as a follow-up to our out-of-band discussion, `typing.Text` was introduced in 3.5.2, so [we should be okay](/wittekm/stone/blob/3cf59644e4ed53f079ca33da29f3733158bc4fa5/tox.ini#L29). 👌

posita: Just thinking out loud here, but are there other places we use `# type: (str)` where we should be using `# type: (typing.Text)`?

braincore: Good point. @wittekm: Could you change these refs as well:
```
$ rg -tpy '\(str\)'
stone/api.py
35:        # type: (str) -> None
41:        # type: (str) -> ApiNamespace
stone/generator.py
118:            target_folder_path (str): Path to the folder where all generated
157:        # type: (str) -> Iterator[None]
217:        # type: (str) -> None
230:        # type: (str) -> None
```

posita: FYI, `# type: (str)` may not play well in Python 2 with `from __future__ import unicode_literals`. Jus' sayin'.

wittekm: Actually mypy understands `unicode_literals` thankfully :)

posita:
@braincore, if you're okay with this, I'll rebase and merge.

wittekm:
@braincore ping 🏓

braincore:
@wittekm: Last thing, could you clean up the commit message?

braincore:
Awesome, thank you!

PR #18, opened 2016-11-20
Fix: Reference to CalcError should be EvalError

akramhussein:
Reference to CalcError should be EvalError or EvalError should be renamed CalcError, otherwise `stone python_types . calc.stone` does not work. 

Outputs:

`calc.stone:3: error: Symbol 'CalcError' is undefined.`


braincore:
Thanks!

PR #19, opened 2016-11-23
Update serializer architecture

posita:
This replaces the function-based serialization code with an extensible object-oriented framework. Older public interfaces are preserved for backward compatibility.

Because the existing `json_encode` interface is now implemented in terms of the new architecture, the existing tests should provide coverage.

### DIFF
--- a/stone/target/python_rsrc/stone_serializers.py
+++ b/stone/target/python_rsrc/stone_serializers.py
@@ -18,17 +18,344 @@ import datetime
 import functools
 import json
 import six
+import typing  # pylint: disable=unused-import
 
 try:
+    from . import stone_base as bb  # pylint: disable=unused-import
     from . import stone_validators as bv
 except (SystemError, ValueError):
     # Catch errors raised when importing a relative module when not in a package.
     # This makes testing this file directly (outside of a package) easier.
+    import stone_validators as bb  # type: ignore # pylint: disable=unused-import
     import stone_validators as bv  # type: ignore
 
+#-------------------------------------------------------------------------
+class StoneEncoderInterface(object):
+    """
+    Interface defining a stone object encoder.
+    """
+
+    def encode(self, validator, value):
+        # type: (bv.Validator, typing.Any) -> typing.Any
+        """
+        Validate ``value`` using ``validator`` and return the encoding.
+
+        Args:
+            validator: the ``stone_validators.Validator`` used to validate
+                ``value``
+            value: the object to encode
+
+        Returns:
+            The encoded object. This is implementation-defined.
+
+        Raises:
+            stone_validators.ValidationError: Raised if ``value`` (or one
+                of its sub-values).
+        """
+        raise NotImplementedError
+
+#-------------------------------------------------------------------------
+class StoneSerializerBase(StoneEncoderInterface):
+
+    def __init__(self, alias_validators=None):
+        # type: (typing.Mapping[bv.Validator, typing.Callable[[typing.Any], None]]) -> None
+        """
+        Constructor, `obviously
+        <http://www.geekalerts.com/ew-hand-sanitizer/>`.
+
+        Args:
+            alias_validators (``typing.Mapping``, optional): A mapping of
+                custom validation callables in the format
+                ``{stone_validators.Validator:
+                typing.Callable[[typing.Any], None], ...}``. These callables must
+                raise a ``stone_validators.ValidationError`` on failure.
+                Defaults to ``None``.
+        """
+        self._alias_validators = {}  # type: typing.Dict[bv.Validator, typing.Callable[[typing.Any], None]]
+
+        if alias_validators is not None:
+            self._alias_validators.update(alias_validators)
+
+    @property
+    def alias_validators(self):
+        """
+        A ``typing.Mapping`` of custom validation callables in the format
+        ``{stone_validators.Validator: typing.Callable[typing.Any],
+        ...}``.
+        """
+        return self._alias_validators
+
+    def encode(self, validator, value):
+        return self.encode_sub(validator, value)
+
+    def encode_sub(self, validator, value):
+        # type: (bv.Validator, typing.Any) -> typing.Any
+        """
+        Callback intended to be called by other ``encode`` methods to
+        delegate encoding of sub-values. Arguments have the same semantics
+        as with the ``encode`` method.
+        """
+        if isinstance(validator, bv.List):
+            # Because Lists are mutable, we always validate them during
+            # serialization
+            validate_f = validator.validate
+            encode_f = self.encode_list
+        elif isinstance(validator, bv.Nullable):
+            validate_f = validator.validate
+            encode_f = self.encode_nullable
+        elif isinstance(validator, bv.Primitive):
+            validate_f = validator.validate
+            encode_f = self.encode_primitive
+        elif isinstance(validator, bv.Struct):
+            if isinstance(validator, bv.StructTree):
+                validate_f = validator.validate
+                encode_f = self.encode_struct_tree
+            else:
+                # Fields are already validated on assignment
+                validate_f = validator.validate_type_only
+                encode_f = self.encode_struct
+        elif isinstance(validator, bv.Union):
+            # Fields are already validated on assignment
+            validate_f = validator.validate_type_only
+            encode_f = self.encode_union
+        else:
+            raise bv.ValidationError('Unsupported data type {}'.format(type(validator).__name__))
+
+        validate_f(value)
+
+        return encode_f(validator, value)
+
+    def encode_list(self, validator, value):
+        # type: (bv.List, typing.Any) -> typing.Any
+        """
+        Callback for serializing a ``stone_validators.List``. Arguments
+        have the same semantics as with the ``encode`` method.
+        """
+        raise NotImplementedError
+
+    def encode_nullable(self, validator, value):
+        # type: (bv.Nullable, typing.Any) -> typing.Any
+        """
+        Callback for serializing a ``stone_validators.Nullable``.
+        Arguments have the same semantics as with the ``encode`` method.
+        """
+        raise NotImplementedError
+
+    def encode_primitive(self, validator, value):
+        # type: (bv.Primitive, typing.Any) -> typing.Any
+        """
+        Callback for serializing a ``stone_validators.Primitive``.
+        Arguments have the same semantics as with the ``encode`` method.
+        """
+        raise NotImplementedError
+
+    def encode_struct(self, validator, value):
+        # type: (bv.Struct, typing.Any) -> typing.Any
+        """
+        Callback for serializing a ``stone_validators.Struct``. Arguments
+        have the same semantics as with the ``encode`` method.
+        """
+        raise NotImplementedError
+
+    def encode_struct_tree(self, validator, value):
+        # type: (bv.StructTree, typing.Any) -> typing.Any
+        """
+        Callback for serializing a ``stone_validators.StructTree``.
+        Arguments have the same semantics as with the ``encode`` method.
+        """
+        raise NotImplementedError
+
+    def encode_union(self, validator, value):
+        # type: (bv.Union, bb.Union) -> typing.Any
+        """
+        Callback for serializing a ``stone_validators.Union``. Arguments
+        have the same semantics as with the ``encode`` method.
+        """
+        raise NotImplementedError
+
+#-------------------------------------------------------------------------
+class StoneToPythonPrimitiveSerializer(StoneSerializerBase):
+
+    def __init__(self, alias_validators=None, for_msgpack=False, old_style=False):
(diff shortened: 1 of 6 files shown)


posita:
Because the existing `json_encode` interface is now implemented in terms of the new architecture, the existing tests should provide coverage.

stone/target/python_rsrc/stone_serializers.py
```diff
@@ -24,11 +24,471 @@
 except (SystemError, ValueError):
     # Catch errors raised when importing a relative module when not in a package.
     # This makes testing this file directly (outside of a package) easier.
-    import stone_validators as bv  # type: ignore
+    import stone_validators as bv # type: ignore
```
braincore: End of line comments are separated from the code by two spaces: https://www.python.org/dev/peps/pep-0008/#inline-comments

stone/target/python_rsrc/stone_serializers.py
```diff
@@ -24,11 +24,471 @@
...
+class StoneSerializerBase(StoneEncoderInterface):
+
+    @staticmethod
+    def out_file_not_implemented(f):
+        """
+        A specialized descriptor class used to decorate various ``encode``
+        methods to assert that the ``out_file`` parameter is ``None``, and
+        raise a ``NotImplementedError`` if it is not. This can be useful
+        where an implementation does not support serializing directly to a
+        file object (e.g., if it's output is not a string).
```
braincore: it's -> its

stone/target/python_rsrc/stone_serializers.py
```diff
@@ -24,11 +24,471 @@
...
+        """
+        return self._alias_validators
+
+    def encode(self, validator, value):
+        return self.encode_sub(validator, value)
+
+    def encode_to_file(self, validator, value, out_file):
+        return self.encode_sub(validator, value, out_file)
+
+    def encode_sub(self, validator, value, out_file=None):
```
braincore: I like this!

test/test_generator.py
```diff
@@ -213,3 +215,6 @@ def test_code_generator_block_gen(self):
     def test_generator_cmdline(self):
         t = TesterCmdline(None, ['-v'])
         self.assertTrue(t.args.verbose)
+
+if __name__ == '__main__':
```
braincore: Oh interesting, you execute the test file directly? I use `pytest`.

stone/target/python_rsrc/stone_serializers.py
```diff
@@ -24,11 +24,471 @@
...
+                Defaults to ``False``.
+        """
+        super(StoneToPythonPrimitiveSerializer, self).__init__(alias_validators=alias_validators)
+        self._for_msgpack = for_msgpack
+        self._old_style = old_style
+
+    @property
+    def for_msgpack(self):
+        """
+        EXPIRIMENTAL: A flag associated with the serializer indicating
```
braincore: EXPIRIMENTAL -> EXPERIMENTAL.

Unless you're trying to tell me that you think we should expire this feature.

stone/target/python_rsrc/stone_serializers.py
```diff
@@ -24,11 +24,471 @@
...
+        """
+        raise NotImplementedError
+
+#-------------------------------------------------------------------------
+class StoneToPythonPrimitiveSerializer(StoneSerializerBase):
+
+    def __init__(self, alias_validators=None, for_msgpack=False, old_style=False):
+        """
+        Args:
+            alias_validators (``collections.Mapping``, optional): Passed
```
braincore: Let's move this to typing.Mapping. Of course, we'll actually need to do a much larger adjustment of all of this to be better PEP-484 citizens: https://www.python.org/dev/peps/pep-0484

stone/target/python_rsrc/stone_serializers.py
```diff
@@ -24,11 +24,471 @@
...
+            encoded object where ``out_file`` is ``None``, and return
+            ``None`` if ``out_file`` is provided (and supported)
+
+        Raises:
+            stone_validators.ValidationError: Raised if ``value`` (or one
+                of its sub-values).
+        """
+        raise NotImplementedError
+
+    def encode_to_file(self, validator, value, out_file):
```
braincore: Any reason why we're adding `encode_to_file`? We're not using it in server code right?

I can imagine a future where we do the encoding in a streaming fashion which would be cool, but this isn't that. It's just a json_dump wrapper which I think is unnecessary. The caller can do that themselves with the response from `encode()`.

stone/target/python_rsrc/stone_serializers.py
```diff
@@ -24,11 +24,471 @@
...
+        json.dump(super(StoneToJsonSerializer, self).encode(validator, value), out_file)
+
+#-------------------------------------------------------------------------
+class StoneSerializerWrapper(object):
+    """
+    Wrapper object for a ``StoneSerializerBase`` to be used as a perverse
+    kind of adapter.
+    """
+
+    @classmethod
```
braincore: Why add a factory system at all? What are you trying to achieve?

stone/target/python_rsrc/stone_serializers.py
posita: Huh. Would’ya lookit that. If yer not careful, ya learn sumthin’ ever’ day! 🤓

stone/target/python_rsrc/stone_serializers.py
```diff
@@ -24,11 +24,471 @@
...
+
+        if alias_validators is not None:
+            self._alias_validators.update(alias_validators)
+
+    @property
+    def alias_validators(self):
+        """
+        A ``collections.Mapping`` of custom validation callables in the
+        format ``{stone_validators.Validator:
+        collections.Callable(object), ...}``.
```
posita: Fixed.

stone/target/python_rsrc/stone_serializers.py
posita: Success!

test/test_generator.py
posita: One can still use `pytest` or whatever. This allows individual file execution without additional dependencies. It's a quick-n-dirty way for anyone to execute a single suite of tests in a particular file. It's not uncommon as far as patterns go (at least from what I've seen in open source projects). It doesn't really add much, but might be familiar to some.

stone/target/python_rsrc/stone_serializers.py
posita: My goodness. What was I smoking when I made this? 🚬 😮  Maybe I need to expire _my_ mental….

stone/target/python_rsrc/stone_serializers.py
posita: Working on it….

stone/target/python_rsrc/stone_serializers.py
posita: > I can imagine a future where we do the encoding in a streaming fashion which would be cool, but this isn't that.

It's true that implementation doesn't exist yet, but without this function, it's not obvious how that would work.

Serialization tends to risk being expensive. The quick-and-dirty (but interface-compatible) stand-in (for now) is to create the intermediate object hierarchy, then serialize it. But this interface allows a different (direct-to-buffer) implementation in the future while maintaining a (somewhat) sensible interface today?

(At least that was the goal. 😊)

stone/target/python_rsrc/stone_serializers.py
posita: Removed as unnecessary cruft. 🗑

stone/target/python_rsrc/stone_serializers.py
posita: Done.

test/test_generator.py
braincore: Works for me!

stone/target/python_rsrc/stone_serializers.py
braincore: I'd be okay with this if we were using it and planned to improve it. But since it isn't being used, and it currently doesn't provide any performance benefit, I think it's unnecessary baggage on the diff. Not feeling that strongly about it though.

stone/target/python_rsrc/stone_serializers.py
```diff
@@ -18,17 +18,449 @@
...
+class StoneToJsonSerializer(StoneToPythonPrimitiveSerializer):
+
+    def encode(self, validator, value):
+        return json.dumps(super(StoneToJsonSerializer, self).encode(validator, value))
+
+    def encode_to_file(self, validator, value, out_file):
+        json.dump(super(StoneToJsonSerializer, self).encode(validator, value), out_file)
+
+#-------------------------------------------------------------------------
+class StoneSerializerWrapper(object):
```
braincore: Is this still necessary?

stone/target/python_rsrc/stone_serializers.py
posita: Okay, I'm sold. The current implementation uses `json.dumps` (as opposed to `json.dump(…, io.BytesIO())`), and `json.dumps` is almost 10× faster (at least according to my tests¹):

```sh
% python …/test.py
running <function mk_f_dump at 0x10f6e9c08> (run 1 of 10)...1.339 s
running <function mk_f_dump at 0x10f6e9c08> (run 2 of 10)...1.478 s
running <function mk_f_dump at 0x10f6e9c08> (run 3 of 10)...1.353 s
running <function mk_f_dump at 0x10f6e9c08> (run 4 of 10)...1.339 s
running <function mk_f_dump at 0x10f6e9c08> (run 5 of 10)...1.334 s
running <function mk_f_dump at 0x10f6e9c08> (run 6 of 10)...1.327 s
running <function mk_f_dump at 0x10f6e9c08> (run 7 of 10)...1.319 s
running <function mk_f_dump at 0x10f6e9c08> (run 8 of 10)...1.308 s
running <function mk_f_dump at 0x10f6e9c08> (run 9 of 10)...1.319 s
running <function mk_f_dump at 0x10f6e9c08> (run 10 of 10)...1.317 s
<function mk_f_dump at 0x10f6e9c08>: 1.343 avg s (10 runs)
running <function mk_f_dumps at 0x10f6e9c80> (run 1 of 10)...0.167 s
running <function mk_f_dumps at 0x10f6e9c80> (run 2 of 10)...0.162 s
running <function mk_f_dumps at 0x10f6e9c80> (run 3 of 10)...0.168 s
running <function mk_f_dumps at 0x10f6e9c80> (run 4 of 10)...0.158 s
running <function mk_f_dumps at 0x10f6e9c80> (run 5 of 10)...0.162 s
running <function mk_f_dumps at 0x10f6e9c80> (run 6 of 10)...0.170 s
running <function mk_f_dumps at 0x10f6e9c80> (run 7 of 10)...0.160 s
running <function mk_f_dumps at 0x10f6e9c80> (run 8 of 10)...0.174 s
running <function mk_f_dumps at 0x10f6e9c80> (run 9 of 10)...0.161 s
running <function mk_f_dumps at 0x10f6e9c80> (run 10 of 10)...0.187 s
<function mk_f_dumps at 0x10f6e9c80>: 0.167 avg s (10 runs)
```

Removed `encode_to_file` and all orphaned `out_file=None` arguments.

---

¹ `test.py` source:

```py
# test.py
from __future__ import absolute_import, division, print_function, unicode_literals

import functools
import io
import json
import time

def mknode(depth):
    d = {}
    if depth > 0:
        for i in range(depth, depth + 10):
            if i & 0x1:
                i = str(i)
            d[i] = mknode(depth - 1)
        for i in range(depth + 10, depth + 20):
            if not i & 0x1:
                i = str(i)
            d[i] = [ mknode(depth - 1) for _ in range(depth) ]
    for i in range(depth + 20, depth + 40):
        if i & 0x1:
            v = str(i)
        else:
            v = i
        if i & 0x2:
            k = str(i)
        else:
            k = i
        d[k] = v
    return d

def time_f(mk_f, n=10, d=3):
    tot = 0
    for i in range(n):
        candidate = mknode(d)
        f = mk_f(candidate)
        print('running {} (run {} of {})...'.format(mk_f, i + 1, n), end='')
        s = time.time()
        f()
        e = time.time()
        t = e - s
        print('{:.3f} s'.format(t))
        tot += t
    print('{}: {:.3f} avg s ({} runs)'.format(mk_f, tot / n, n))

def mk_f_dump(obj):
    fp = io.BytesIO()
    return functools.partial(json.dump, obj, fp)

def mk_f_dumps(obj):
    return functools.partial(json.dumps, obj)

for mk_f in mk_f_dump, mk_f_dumps:
    time_f(mk_f)
```

stone/target/python_rsrc/stone_serializers.py
posita: Removed.

braincore:
Cool! Can you squash and finalize the log?

posita:
Done.

screen shot 2016-11-29 at 17 51 43

PR #28, opened 2016-12-09
READY FOR REVIEW - Reintroduce composite typing

posita:
This reintroduces some of the composite typing (but properly this time) that was removed as part of 09a31fd2b158d91c429da4efdb7d4d510d572465. These were originally removed as work-arounds for mypy validation.


tox.ini
```diff
@@ -41,7 +41,7 @@ commands =
     python setup.py check -m -r -s -v
 
 deps =
-    docutils
+    docutils==0.12
```
posita: `docutils` _just_ released `0.13.1` which caused our builds to fail inexplicably (see, e.g., [this failing build](https://travis-ci.org/dropbox/stone/jobs/182914200) vs. [this passing one](https://travis-ci.org/dropbox/stone/jobs/182914480), where the only difference is that the latter pins `docutils` to `0.12`). I'm pretty sure this is a bug in `docutils` (it appears on Travis CI in both Python 2.7.9 and 3.4.2, but not on my local machine with Python 2.7.10 and Python 3.5.2). I will follow up with the maintainers of that project. However, until a fix is released, this should remain pinned.

.gitignore
```diff
@@ -1,17 +1,20 @@
...
-*.pyo
-*.swp
-parser.out
-parsetab.py
-.idea
+/.cache
 /.tox
+/.tox-*
+/.venv
+/.venv-*
```
posita: I added the above three entries. The other changes to this file are just a reordering of existing entries.

LICENSE
```diff
@@ -1,20 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
```
posita: This was a misguided effort to troubleshoot our failing builds, which _looked_ like a reStructuredText error, but was really a problem with an updated buggy `docutils`. (See below.) Technically, this change isn't required, but I prefer this.

README.rst
```diff
@@ -1,6 +1,6 @@
-*****
+=====
 Stone
-*****
+=====
```
posita: Another misguided attempt to troubleshoot the `docutils` update problem, but I _think_ this is more idiomatic. (See http://docutils.sourceforge.net/docs/user/rst/quickstart.html#document-title-subtitle.)

mypy-run.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/bash -eux
 
-EXCLUDE="(^example|ez_setup.py)"
+EXCLUDE='(^example/|^ez_setup\.py$)'
```
posita: This didn't hurt anything, but it was technically incorrect.

mypy-run.sh
```diff
@@ -1,9 +1,9 @@
...
 
 # Include all Python files registered in Git, that don't occur in $EXCLUDE.
-INCLUDE=$(git ls-files "$@" | grep '\.py$' | egrep -v "$EXCLUDE" | tr '\n' '\0' | xargs -0 | cat)
-MYPY_CMD="mypy --silent-imports --fast-parser"
```
posita: Config now lives in `mypy.ini`.

setup.py
```diff
@@ -33,7 +30,7 @@
     maintainer_email='dev-platform@dropbox.com',
     url='https://github.com/dropbox/stone',
     install_requires=install_reqs,
-    license=LICENSE,
+    license=open('LICENSE').read(),
```
posita: Technically, this should probably be a string like "MIT" (if that's what our license is). If it is a known license, we should probably add a corresponding `License :: …` [classifier](https://pypi.python.org/pypi?%3Aaction=list_classifiers).

stone/api.py
```diff
@@ -18,19 +19,21 @@
...
-    NamespaceDict = typing.Dict[str, b'ApiNamespace']
+__MYPY = False
+if __MYPY:
+    # TODO: This can be changed back to a single declaration with a
+    # unicode literal after <https://github.com/python/mypy/pull/2516>
+    # makes it into a PyPi release
+    if six.PY3:
+        NamespaceDict = typing.Dict[str, 'ApiNamespace']
+    else:
+        NamespaceDict = typing.Dict[str, b'ApiNamespace']
```
posita: This is a common hack to allow running of Stone code even if the `typing` module is sub-par. It should be possible to extend this such that one could remove the dependency on `typing` altogether, but that's probably the subject of another PR.

stone/generator.py
```diff
@@ -24,9 +24,12 @@
...
 
-# Generic Dict key-val types
-K = typing.TypeVar('K')
-V = typing.TypeVar('V')
+__MYPY = False
+if __MYPY:
+    # Generic Dict key-val types
+    DelimTuple = typing.Tuple[typing.Text, typing.Text]
+    K = typing.TypeVar('K')
+    V = typing.TypeVar('V')
```
posita: Same hack as above.

setup.py
```diff
@@ -33,20 +30,21 @@
     maintainer_email='dev-platform@dropbox.com',
     url='https://github.com/dropbox/stone',
     install_requires=install_reqs,
-    license=LICENSE,
+    license=open('LICENSE'),
```
posita: Apparently both `license` and `long_description` [take file objects as arguments](http://setuptools.readthedocs.io/en/latest/setuptools.html#metadata). Who knew?

.gitignore
```diff
@@ -1,17 +1,20 @@
...
+/stone.egg-info
 .arcconfig
+.DS_Store
+.idea
+parser.out
+parsetab.py
+*.pyc
+*.pyo
+*.swp
+*~
```
braincore: Random q: Why would you put these down there?

tox.ini
posita: UPDATE: I filed this as [`docutils` bug #302](https://sourceforge.net/p/docutils/bugs/302/).

.gitignore
braincore: Haha interesting. I'd always go least-specific to most. So I guess you're likely to prepend to your list, and I'm likely to append.

LICENSE
braincore: I'd say we should keep it as is only because it's a standard copy of our license (including the line breaks). It'll be easier to see that they're the same agreement using diff.

setup.py
```diff
@@ -47,6 +44,7 @@
     classifiers=[
         'Development Status :: 4 - Beta',
         'Intended Audience :: Developers',
+        'License :: OSI Approved :: MIT License',
```
braincore: Is this kosher? We're technically under the Dropbox License which isn't far off from I think the Apache license.

lfaraone: The work is under the ["MIT" license](https://opensource.org/licenses/MIT) (more precisely, Expat), so this is correct.

stone/api.py
posita: `__MYPY` should never be set to `True`. This is a hack to confuse mypy into passing without blowing up in environments running the actual library code where `typing` is the wrong version (and doesn't support something we want).

.gitignore
posita: I can revert or change it if desired. I don't think the list is long enough to consider pattern matching performance within `git` (if list order is even important, which I don't know).

stone/api.py
posita: Actually, we may want to consider guarding `import typing` with this hack as well, to avoid `typing` as a requirement completely.

setup.py
posita: No license this short will preserve patent rights by those who use or derive works from the software. Most OSS licenses implicitly grant (and some explicitly permit) non-exclusive practice rights.¹ 

--

¹ This is not intended as legal advice. 😉

tox.ini
posita: UPDATE: `docutils` maintainers do not appear interested in helping without additional information from our end. They are [blaming it on our application](https://sourceforge.net/p/docutils/bugs/302/#7107).

braincore:
This looks good to me. Any more TODOs?

posita:
> This looks good to me. Any more TODOs?

Nope. Just an FYI: I reverted my `LICENSE` reformatting in response to the comments. Other than that, I don't think there's anything that should be done within the scope of this PR. I will track down the `docutils` SNAFU separately.

Is it okay if I rebase and merge?

PR #30, opened 2016-12-14
READY FOR REVIEW - Remove stale entries from .gitignore

posita:
Several entries are no longer needed in `.gitignore`. This removes them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,14 +7,11 @@
 /build
 /dist
 /MANIFEST
-/setuptools-3.1-py2.7.egg
-/setuptools-3.1.zip
 /stone.egg-info
 .arcconfig
 .DS_Store
 .idea
 parser.out
-parsetab.py
 *.pyc
 *.pyo
 *.swp


.gitignore
```diff
@@ -6,14 +6,10 @@
 /build
 /dist
 /MANIFEST
-/setuptools-3.1-py2.7.egg
-/setuptools-3.1.zip
 /stone.egg-info
 .arcconfig
 .DS_Store
 .idea
-parser.out
```
posita: Is that something that a Stone developer might create? Should we be automating tests or other checks that invoke those things?

braincore: Ya, a bit hazy, but a stone developer might create the file just by opening a python console and doing a relative import of the stone parser.

posita: Is that a bug?

braincore: `ply`'s a weird library all around. For example, you specify grammar rules in method docstrings. It's simply the libraries behavior to dump intermediate files into the cwd.

posita: Okay, I can't repro, but I added it back in.

braincore: I think I remembered another detail. Did you run stone with `-v` or `-vv`? Depending on the level of verbosity I think it changes the debug flag passed into `ply`.

PR #31, opened 2016-12-14
READY FOR REVIEW - Remove typing as runtime dependency

posita:
DEPENDS ON #37. This removes `typing` as a runtime dependency of Stone so that Python 2.7 users do not have to install it unnecessarily. It is still a build/test dependency for those who chose to use take advantage of it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /.cache
+/.eggs
 /.tox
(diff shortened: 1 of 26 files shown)


braincore:
Using `typing` in PY2 is so sad compared to PY3.

PR #35, opened 2017-01-18
WIP - Allow for relaxed validation during encoding

posita:
Allow for relaxed validation during encoding to accommodate situations where a partially-encoded value is preferable to raising a `ValidationError` exception (e.g., for logging purposes).


posita:
Abandoning strategy.

PR #36, opened 2017-01-19
READY FOR REVIEW - Handle encoding dates with years < 1900 on Python 2

posita:
This uses an alternate `strftime` implementation to overcome a Python 2 limitation for supporting years < 1900. Python 3 does not suffer from this issue.

### DIFF
--- a/stone/target/python_rsrc/stone_serializers.py
+++ b/stone/target/python_rsrc/stone_serializers.py
@@ -17,7 +17,9 @@ import collections
 import datetime
 import functools
 import json
+import re
 import six
+import time
 import typing  # noqa: F401 # pylint: disable=unused-import
 
 try:
@@ -229,7 +231,7 @@ class StoneToPythonPrimitiveSerializer(StoneSerializerBase):
         if isinstance(validator, bv.Void):
             return None
         elif isinstance(validator, bv.Timestamp):
-            return value.strftime(validator.format)
+            return _strftime(value, validator.format)
         elif isinstance(validator, bv.Bytes):
             if self.for_msgpack:
                 return value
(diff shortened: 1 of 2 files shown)


stone/target/python_rsrc/stone_serializers.py
```diff
@@ -226,23 +226,27 @@ def encode_primitive(self, validator, value):
...
-        elif isinstance(validator, bv.Integer) \
-                    and isinstance(value, bool):
-            # bool is sub-class of int so it passes Integer validation,
-            # but we want the bool to be encoded as ``0`` or ``1``, rather
-            # than ``False`` or ``True``, respectively
-            return int(value)
-        else:
-            return value
+                return value
+        except Exception as exc:
```
cakoose: What exceptions do you expect here?  Would it be enough to limit the `try` to the `b64encode` call?

braincore:
Was there a specific exception that wasn't being caught and re-raised as a `ValidationError`? Currently, I've tried to make sure that every possible error is caught and re-raised as a `ValidationError` with an appropriate message. The message is important because those are returned in the body of `400 Bad Request` responses that are triggered.

stone/target/python_rsrc/stone_serializers.py
posita: Not really. Depending on what `value` was, there's quite a bit in theory that could go wrong here. For example, `_strftime` could raise a `TypeError` or `ValueError` (or even an `AttributeError` if it ended up not being a `datetime` object. `base64.b64encode(value)` could raise a `TypeError` or `ValueError`.

posita:
I've removed the exception re-casting. All that remains is the `strtime` fix.

stone/target/python_rsrc/stone_serializers.py
```diff
@@ -229,7 +231,7 @@ def encode_primitive(self, validator, value):
         if isinstance(validator, bv.Void):
             return None
         elif isinstance(validator, bv.Timestamp):
-            return value.strftime(validator.format)
+            return _strftime(value, validator.format)
```
braincore: Does it make sense to use this custom `strftime` only when we're in PY2?

test/test_python_gen.py
```diff
@@ -1339,5 +1344,38 @@ def test_struct_union_default(self):
         assert s.u == self.ns2.BaseU.z
 
 
+# Adapted from:
+# http://code.activestate.com/recipes/306860-proleptic-gregorian-dates-and-strftime-before-1900/
+# Make sure that the day names are in order from 0001/01/01 until
+# 2000/08/01
+def test_strftime():
```
braincore: Is this test executed even when it isn't part of the `unittest.TestCase` subclass?

stone/target/python_rsrc/stone_serializers.py
posita: I think so. Python 3 does not suffer from this limitation.

From https://docs.python.org/2/library/datetime.html#strftime-and-strptime-behavior:

> The exact range of years for which `strftime()` works also varies across platforms. Regardless of platform, years before 1900 cannot be used.

From https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior:

> 2. The `strptime()` method can parse years in the full [1, 9999] range, but years < 1000 must be zero-filled to 4-digit width.
>
>     _Changed in version 3.2:_ In previous versions, `strftime()` method was restricted to years >= 1900.
>
>     _Changed in version 3.3:_ In version 3.2, `strftime()` method was restricted to years >= 1000.

Cf. http://bugs.python.org/issue1777412 (confirming that support of years < 1900 is a _feature_, not a bug, and would only be added to Python 3).

test/test_python_gen.py
posita: That's a good point. It _is_ picked up by tox's test discovery, but it does _not_ get run by `unittest.main()`. I'll wrap it in a `unittest.TestCase`.

stone/target/python_rsrc/stone_serializers.py
```diff
@@ -853,6 +855,73 @@ def _make_stone_friendly(
...
+
+    return sites
+
+# Every 28 years the calendar repeats, except through century leap years
+# where it's 6 years. But only if you're using the Gregorian calendar. ;)
+def _strftime(dt, fmt):
+    try:
+        return dt.strftime(fmt)
+    except ValueError:
+        if six.PY3 or dt.year > 1900:
```
posita: @braincore, I'm sorry I misunderstood your previous comment. Is this what you were referring to? Basically, this tries to use the library function first. If we're executing in Python 3, or if the year is valid, then re-`raise` any caught `ValueError`. Otherwise, we're in Python 2 and the year is something we can compensate for.

If I was extra paranoid, I might try to look at the error string of the `ValueError` just to make sure it wasn't due to something else (even though the year is less than 1900).

braincore:
lgtm (need minor PY4 future proofing haha), feel free to rebase when ready.

posita:
> lgtm (need minor PY4 future proofing haha), feel free to rebase when ready.

Fixed, but I'm going to hold you to lobbying for a PY3-compatible `datetime.strftime` in Python 4. :wink:

PR #37, opened 2017-01-23
READY FOR REVIEW - Fix type issue

posita:


stone/target/tsd_types.py
```diff
@@ -112,7 +112,8 @@ class TSDTypesGenerator(CodeGenerator):
 
     preserve_aliases = True
 
-    cur_namespace = None
+    # Instance var of the current namespace being generated
+    cur_namespace = None  # type: ignore # TODO: What is the type of this thing?!
```
braincore: It's a `stone.api.ApiNamespace` object.

stone/target/js_client.py
```diff
@@ -44,7 +45,7 @@ class JavascriptClientGenerator(CodeGenerator):
     cmdline_parser = _cmdline_parser
 
     # Instance var of the current namespace being generated
-    cur_namespace = None  # type: ignore # TODO: What is the type of this thing?!
+    cur_namespace = None  # type: ApiNamespace
```
braincore: Does this need to be marked `Optional[ApiNamespace]` or does mypy infer that from the declaration?

posita: Probably yes, even though our current mypy tests pass (although I can't quite figure out why this slips by).

posita: Fixed now.

PR #38, opened 2017-01-25
READY FOR REVIEW - Add ISO-8601 timestamp handling

posita:
This adds `python-dateutil` and `pytz` _only_ as testing dependencies.


stone/target/python_rsrc/stone_serializers.py
```diff
@@ -232,6 +256,8 @@ def encode_primitive(self, validator, value):
             return None
         elif isinstance(validator, bv.Timestamp):
             return _strftime(value, validator.format)
+        elif isinstance(validator, bv.Iso8601Timestamp):
+            return _strftime(value.astimezone(TZINFO_UTC), validator.output_format)
```
posita: This is to support the case where the `datetime` object we're provided is in a different timezone. This basically imposes that all encoded timestamps are always in UTC. This is basically so format strings with hard-coded timezone values (e.g., `%Y-%m-%dT%H:%M:%SZ`, `%Y-%m-%dT%H:%M:%S +0000`) will work. You can still trick it by lying (e.g., `%Y-%m-%dT%H:%M:%S +1234`), but the former (anti)pattern is common enough where I think it makes sense.

stone/data_type.py
```diff
@@ -444,6 +449,49 @@ def check_attr_repr(self, attr_field):
...
+            if isinstance(msg, six.binary_type):
+                # For Python 2 compatibility.
+                msg = msg.decode('utf-8')
+            raise InvalidSpec(msg, attr_field.lineno, attr_field.path)
+
+    def _parse(self, val):
+        if self.input_re is not None and not self.input_re.search(val):
+            raise ValueError('{} does not match {}'.format(val, self.input_re.pattern))
+        # Raises a TypeError if val cannot be parsed
+        return dateutil.parser.parse(val)
```
braincore: It isn't great that our standard for what is valid for this type is simply what dateutil can parse.

braincore:
I'm feeling a little bit uncomfortable with this one. I think the major concern I have is with regards to portability across languages. At this point, we're depending on dateutil as the measure of what passes as Iso8601. I don't know whether depending on dateutil is a good choice nor do I know whether other languages have something similar.

What if Timestamp could simply take a list of valid formats that the deserializer will try in order?

stone/data_type.py
posita: I supposed we could add a default regex? The test case [has one](https://github.com/posita/stone/blob/4ee632e288c5c4beaa4595379c8d05f0dfd23d34/test/test_python_gen.py#L138), but it's a bit complex.

posita:
Are you worried that `dateutil` tries too hard to parse timestamps in weird formats (i.e., those which are outside the ISO 8601 specification)? Would you feel better if there existed a library that strictly limited its parsing to ISO 8601 formats?

> What if Timestamp could simply take a list of valid formats that the deserializer will try in order?

🤢

braincore:
Ya. If every language (we plan to support) had an accurate (ideally builtin though we could add dependencies) ISO 8601 parser, I'd be more convinced. Then we could more honestly say it's an ISO 8601 timestamp type, and we could expect consistent behavior across languages.

posita:
Various validation regexes (although I'm not sure that's the way we want to go; they're computationally expensive):

- http://stackoverflow.com/questions/28020805/regex-validate-correct-iso8601-date-string-with-time#28022901
- https://www.safaribooksonline.com/library/view/regular-expressions-cookbook/9781449327453/ch04s07.html
- http://www.pelagodesign.com/blog/2009/05/20/iso-8601-date-validation-that-doesnt-suck/

posita:
> Ya. If every language (we plan to support) had an accurate (ideally builtin though we could add dependencies) ISO 8601 parser, I'd be more convinced. Then we could more honestly say it's an ISO 8601 timestamp type, and we could expect consistent behavior across languages.

I'm not sure I understand. I _think_ `dateutil` strives to be a parser that includes ISO 8601 parsing capabilities. In other words, if its advertising is accurate, it's a superset. Also, `dateutil` should only implicated during deserialization.

braincore:
One of the (unwritten) development principles has been that Stone should be capable of generating server-side validators in any of its target languages, and provide consistent deserialization requirements across them. Currently, we only use Stone for our Python webserver. But, we could start porting server-side API endpoints to other languages at Dropbox used for serving such as JS (node) or Go.

So the problem is that `dateutil` unlike `strptime()` doesn't have a portable implementation. Our Go, JS, and Python timestamp deserializers would all behave differently.

posita:
> One of the (unwritten) development principles has been that Stone should be capable of generating server-side validators in any of its target languages, and provide consistent deserialization requirements across them.

Okay, this is what I was missing. I understand now.

posita:
FWIW, I don't know if that's a practical design goal. There's so much variation on server-side capabilities.

braincore:
Maybe, but this is also just the serialization/deserialization component. AFAIK, our current feature set is consistently enforceable in other languages.

posita:
> Maybe, but this is also just the serialization/deserialization component. AFAIK, our current feature set is consistently enforceable in other languages.

FWIW, `strptime`/`strftime` is kind of false standard. The ANSI C standard is insufficient (without OS-specific extensions) to support many common ISO 8601 formats.

After taking a look at [`dateutil`'s source](https://github.com/dateutil/dateutil/blob/master/dateutil/parser.py), I'm convinced that with the default of `fuzzy=False`, it's going to do the right (or at least reasonable) thing within the constraints of most standards (ISO 8601, W3C, etc.).

I don't think we have to have identical server-side behavior. Actually, I think this is untenable (at least with the current state of the world). If we're being formal about it, we could, for example, declare:

- Server-side implementations of `Iso8601Timestamp`**MUST** support deserialization of ISO 8601 (2004) compliant text strings into a timezone-aware native data structure
- Server-side implementations of `Iso8601Timestamp` **SHOULD** prefer deserialization into standard library data structures or timestamp implementations
- Server-side implementations of `Iso8601Timestamp` **MAY** support deserialization of additional formats or specifications in addition to ISO 8601 (2004) compliant text strings
- Server-side implementations of `Iso8601Timestamp` **MUST** interpret ISO 8601 (2004) compliant text strings according to ISO 8601 (2004), even where there is an ambiguity with another supported format or specification

posita:
Abandoned in favor of #52.

PR #40, opened 2017-02-15
READY FOR REVIEW - Fix `typed-ast` dependency problem for MyPy checks

posita:
Future TODO: Unpin `typed-ast` with next MyPy release.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -28,7 +28,8 @@ commands =
 
 deps =
     mypy-lang
-    typed_ast
+    typed-ast < 1.0.0
+    # git+git://github.com/python/mypy.git@3cc52013c35dadda57964c7d4ae6c664bfc1329c
 
 usedevelop = true
 
--- a/stone/api.py
+++ b/stone/api.py
@@ -56,7 +56,7 @@ class Api(object):
         """
         if name not in self.namespaces:
             self.namespaces[name] = ApiNamespace(name)
-        return self.namespaces.get(name)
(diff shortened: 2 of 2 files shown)


tox.ini
```diff
@@ -28,7 +28,8 @@ commands =
 
 deps =
     mypy-lang
-    typed_ast
+    typed-ast < 1.0.0
+    # git+git://github.com/python/mypy.git@3cc52013c35dadda57964c7d4ae6c664bfc1329c
```
posita: There's two possible strategies here:

1. Constrain `typed-ast` to a version that doesn't break the current release of MyPy; or
1. Pin MyPy to a not-yet-released commit that accommodates (and constrains) `typed-ast`.

This uses the first approach. Either way, there's work to undo the pin with the next release of MyPy.

wittekm:
One random note: I see that the underscore was switched to a hyphen. Are you sure that <1 is needed?

posita:
> One random note: I see that the underscore was switched to a hyphen. Are you sure that <1 is needed?

Not needed, but technically, `typed_ast` is not the package name. It ultimately gets mapped to the [canonical `typed-ast`](https://pypi.python.org/pypi/typed-ast/1.0.0). Also, `typed-ast` matches the [corrective instruction from MyPy](https://travis-ci.org/dropbox/stone/jobs/201700439#L196). (Just my OCD rearing its ugly head. It was fine as-was.)

PR #41, opened 2017-02-18
READY FOR REVIEW - Add test-only requirements

posita:
This attempts to isolate any requirements needed solely for testing in a Tox-friendly way.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,6 +27,12 @@ test_reqs = [
     'pytest',
 ]
 
+# WARNING: This imposes limitations on test/requirements.txt such that the
+# full Pip syntax is not supported. See also
+# <http://stackoverflow.com/questions/14399534/>.
+with open('test/requirements.txt') as f:
+    test_reqs += f.read().splitlines()
+
 with open('LICENSE') as f:
     LICENSE = f.read()
 
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,3 +2,4 @@ include ez_setup.py
 include LICENSE
 include *.rst
 include stone/target/swift_rsrc/*
+include test/requirements.txt
--- a/tox.ini
+++ b/tox.ini
@@ -45,6 +45,9 @@ commands =
 deps =
     flake8
     pylint
+    # This probably breaks on Windows. See
+    # <https://github.com/tox-dev/tox/issues/384>.
+    -rtest/requirements.txt
 
 usedevelop = true
 


setup.py
```diff
@@ -2,6 +2,7 @@
 # where package_data is expected to be str and not unicode.
 from __future__ import absolute_import, division, print_function
 
+import os.path
```
braincore: Just `import os`, path is already auto-imported by it. Technically safer this way, but non-standard IMO.

setup.py
```diff
@@ -27,6 +28,12 @@
     'pytest',
 ]
 
+# WARNING: This imposes limitations on test/requirements.txt such that the
+# full Pip syntax is not supported. See also
+# <http://stackoverflow.com/questions/14399534/>.
+with open(os.path.join('test', 'requirements.txt')) as f:
```
braincore: Nit: `with open('test/requirements.txt') as f:`. Works cross platform with Window paths if that's what you were worried about.

braincore:
Feel free to rebase it in yourself once you've addressed nits.

setup.py
posita: Okay, just for the sake of my own edification, _where_ (tf) is that documented? 😳

setup.py
posita: I had no idea this was [as controversial as it is](http://stackoverflow.com/questions/2724348/). I'm fine with changing it, but I cringe thinking about all the times in the past that I've done `import os.path`. 😖

setup.py
posita: Fixed, but for my edification, _where_ (tf) is that documented? 😳

setup.py
braincore: Haha interesting thread. Looks like it's not so obvious what to do :)

I'm guessing the fact that `os` injects `path` into `sys.modules` (and not done as a conditional import...) is a legacy artifact.

setup.py
braincore: haha shoot, i can't find anything that documents that behavior and now i'm worried i'm misremembering.

posita: I think it does work, but I can't figure out how. :man_shrugging: 

The closest thing I found was [`os.path.normpath`](https://docs.python.org/2/library/os.path.html#os.path.normpath) whose documentation states:

>  Windows, it converts forward slashes to backward slashes.

Note there's [`io.open`](https://github.com/python/cpython/blob/0f48ecddfb79ef541500ecfcbda0b4178af1cc29/Modules/_io/_iomodule.c#L288), but it's not the same as the built-in `open`:

```py
Python 2.7.10 (default, Jul 30 2016, 19:40:32)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.34)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import io
>>> io.open
<built-in function open>
>>> open
<built-in function open>
>>> io.open is open
False
```

Neither say anything explicit about calling `os.path.normpath`, and I can't find it in their sources. Even the [Python implementation of `io.open`](https://github.com/python/cpython/blob/0f48ecddfb79ef541500ecfcbda0b4178af1cc29/Lib/_pyio.py#L44) is silent on the issue.

PR #47, opened 2017-04-20
Added private field support.

scobbe:


posita:
Forgive my ignorance, but is this a generally-available feature? Should any of the docs be updated to explain what it is and how to use it?

scobbe:
@posita: This will soon be a public feature. You're correct, the docs should be updated. I will put out an update to this PR to update the docs.

PR #50, opened 2017-06-15
add Map/Dict data type

risforrob:
initial add of work in progress to start reviewing framework changes while I contnue working on tests.

Remaining items to do:
- [x] update README
- [x] finish writing tests.


cakoose:
Since JSON object keys can only be strings, what's the plan for using other types as keys?

Some options:
1. Map keys must be strings.
2. Map keys must be strings or things that can easily be encoded as strings (e.g. numbers can be quoted).
3. Map keys can to be anything, but use a list-of-pairs representation for those.

For example:
```
-- Type --         -- Example JSON value --
Map(String,...)    {"a": ..., "b": ...}
Map(Integer,...)   {"1": ..., "2": ...}  OR  [[1, ...], [2, ...]]
Map(Boolean, ...)   {"true": ..., "false": ...}  OR  [[true, ...], [false, ...]]
Map(Point, ...)    [[{"x": 1, "y": 2}, ...], [{"x": 3, "x": 4}, ...]]

**Comment by risforrob** [comment]
The plan for now is that only strings may be used as keys.  (This would have been clearer if I'd actually updated the documentation before starting this PR).

File: stone/data_type.py [hunk 1]
```diff
@@ -488,6 +488,35 @@ def _check_list_container(self, val):
...
+            raise ParameterError("Only String primitives are supported as key types.")
+
+        self.key_data_type = key_data_type
+        self.value_data_type = value_data_type
+
+    def check(self, val):
+        raise NotImplementedError
+
+    def check_example(self, ex_field):
+        for k,v in ex_field.value.items():
```
posita: It looks like tests pass in CI. 👍 It's just failing on linter stuff. E.G., from https://travis-ci.org/dropbox/stone/jobs/243434091#L285: `stone/data_type.py:505:14: E231 missing whitespace after ','`.

stone/lang/parser.py
```diff
@@ -982,6 +983,46 @@ def p_ex_list_items_extend(self, p):
         p[0].append(p[3])
 
     # --------------------------------------------------------------
+    # Maps
+    #
+
+    def p_ex_map(self, p):
+        """ex_map : LBRACE ex_map_pairs RBRACE
+                  | LBRACE empty RBRACE"""
+        p[0] = {} if p[2] is None else p[2]
```
risforrob: Correct, `ex_map_pairs` should never be an empty list.  The second `empty` case is to handle if somebody writes:
```
struct S
    arg Map(String, String)
    
    example
        arg = {}
````

File: stone/lang/parser.py [hunk 3]
```diff
@@ -982,6 +983,46 @@ def p_ex_list_items_extend(self, p):
...
+
+    def p_ex_map_elem_primitive(self, p):
+        """ex_map_elem : primitive"""
+        p[0] = None if p[1] == StoneNull else p[1]
+
+    def p_ex_map_elem_composit(self, p):
+        """ex_map_elem : ex_map
+                       | ex_list"""
+        p[0] = p[1]
```
risforrob: My understanding was that by deferring the conversion (at least in the `ex_list` case), then I'm not duplicating conversion code in the parser.

scobbe:
This looks good to me as well!

risforrob:
@posita When you have a moment to re-review, this PR is ready to go.

stone/lang/parser.py
posita: Cool. As long as it gracefully handles the cases where it starts out as an `ex_list` (which I don't even know is possible).

posita:
**SHIP IT!**

screen-shot-2013-03-27-at-9 52 12-am

stone/lang/parser.py
risforrob: The following would be considered valid syntax:
```
struct S
    my_arg Map(String, String)

    example default
        my_arg = {[1, 2]: "foo"}
```
but fails validation for the example and thus fails compilation.
```
stone.lang.exception.InvalidSpec: InvalidSpec(u'[1, 2] is an invalid hash key because it cannot be hashed.', 9, '_.stone')
```

posita:
Note: I squashed commits _after_ merging this (requiring rewriting commit history on `master` 😱). The squashed commit for this PR is f3cad0acf2b683224817dc4365331647194b7fe6.

PR #51, opened 2017-07-06
include lists in make_compact

jlaz91:
fixes issue where a list of parameters doesn’t exclude unnecessary lone
.tag keys when generating examples


scobbe:
@jlaz91: Thanks for making this change. Would you mind including our newly-supported Map type as well in this `make_compact` logic?

scobbe:
Thanks!

PR #54, opened 2017-09-12
READY FOR REVIEW - Fix build error (`mypy-lang` -> `mypy`)

posita:
This is a quick fix designed to unbreak builds. More scrutiny may be needed in the affected areas and should be handled in separate PRs. See my comments below regarding the individual changes for details.

### DIFF
--- a/mypy-run.sh
+++ b/mypy-run.sh
@@ -1,6 +1,6 @@
 #!/bin/bash -eux
 
-EXCLUDE='(^example/|^ez_setup\.py$)'
+EXCLUDE='(^example/|^ez_setup\.py$|^setup\.py$)'
 
 # Include all Python files registered in Git, that don't occur in $EXCLUDE.
 INCLUDE=$(git ls-files "$@" | grep '\.py$' | grep -Ev "$EXCLUDE" | tr '\n' '\0' | xargs -0 | cat)
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,5 +1,5 @@
 [mypy]
-fast_parser = True
+# follow_imports = False
+ignore_missing_imports = True
 show_none_errors = True
-silent_imports = True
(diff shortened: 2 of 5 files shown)


mypy-run.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash -eux
 
-EXCLUDE='(^example/|^ez_setup\.py$)'
+EXCLUDE='(^example/|^ez_setup\.py$|^setup\.py$)'
```
posita: `setup.py` imports `ez_setup.py`, so we're explicitly excluding it now. If we want Mypy checking on `setup.py`, we'll have to figure out a way to re-include it and exclude consideration of `ez_setup.py`, because that's _never_ going to pass validation.

stone/backend.py
```diff
@@ -292,7 +292,7 @@ def emit_wrapped_text(
 
     @classmethod
     def process_doc(cls, doc, handler):
-        # type: (str, typing.Callable[[str, str], str]) -> str
+        # type: (str, typing.Callable[[str, str], str]) -> typing.Text
```
posita: I'm pretty sure this necessary to avoid error because `stone/backend.py` has `from __future__ import … unicode_literals`. (See below.)

mypy.ini
```diff
@@ -1,5 +1,5 @@
 [mypy]
-fast_parser = True
+# follow_imports = False
```
posita: `silent_imports` has been split into two separate configurable behaviors:

- `follow_imports`
- `ignore_missing_imports`

I thought setting both to `True` would avoid having to exclude `setup.py` (see comment above) from Mypy validation, but that didn't turn out to be the case. I'm not sure why.

scobbe:
Looks good – thanks @posita!

PR #57, opened 2017-10-30
Add support to generate typescript definitions per namespace into separate file.

svpranay:
The current implementation of typescript backend generates a single monolithic file containing all the namespaces defined in the specs. This change lets users to generate one file per namespace containing its definitions.

Changes : 
1. Modified the `filename` parameter to be optional. If filename is provided, then all the namespace definitions are written to the file with the given filename. Otherwise, one file (with the namespace name) is generated per namespace, which contains the definitions of the namespace.

2. Added a new boolean param : `exclude_error_types`. This flag lets users to optionally insert error types into the definitions file. This has been added to provide an option to avoid including error types to each namespace file when multiple files are generated. 

3. Tests have been added to tsd types generator.


svpranay:
@scobbe Please take a look.

svpranay:
@braincore @risforrob Would you be able to review this PR ? Thanks.

scobbe:
Looks good to me @svpranay! Thanks for making these changes.

PR #59, opened 2017-11-11
Change typescript types output format to declaration files.

svpranay:
Modify the tsd backend to emit declaration files for namespaces instead of typescript modules.

The changes made to convert ts modules to declarations  are : 
- Change the file extension from `.ts` to `.d.ts`.
- A top level `declare` prefix for declaration files.
- `export` all type definitions.
- Change the import style in declaration files. Now uses reference path.
- Avoid emitting empty declaration files. (Added tests.)
- Ensure Timestamp has a type definition in all declaration files for completeness.


svpranay:
@scobbe @qimingyuan

PR #66, opened 2017-12-29
fix determinism bug in _generate_struct_class_reflection_attributes

benjaminp:
Don't rely on set iteration order.


scobbe:
Thanks @benjaminp for fixing this.

PR #69, opened 2018-02-07
correct a code generation determinism bug in _generate_union_class_reflection_attributes

benjaminp:
A very similar story to 9754a09e527ee86e6a1baad6ee2d0a8d12fb465e.


scobbe:
Looks like some unrelated lint errors caused the tox tests to fail. Let me fix them. Thanks for making this change.

PR #70, opened 2018-02-12
Use more friendly name for unexpected booleans

dirn:
When a field receives a boolean but expects another type, the Python
validator will try to show a user-friendly name. While `bool` is a
subclass of `int`, telling the end user that an integer was received can
lead to confusion, especially if they don't know this fact. By
explicitly checking if a value is a boolean before checking for numeric
types, `generic_type_name` can return a more explicit type name.

This will add an `isinstance` check that checks for booleans before
checking for numeric types. The check is being added before the other
type checks so that it isn't captured by the check for
`numbers.Integral`.


dirn:
Done.

PR #73, opened 2018-02-21
Change python_type_stubs generator to use `bv` instead of `stone_validators` (oops!)

wittekm:
﻿a) I was previously annotating as `stone_validators` - but the import actually names them `stone_validators as bv`.

b) I changed the routes and validators to use the [PEP 526](https://www.python.org/dev/peps/pep-0526/)-style variable annotation. Per ilevkivskyi on mypy channel at dropbox's slack, this is the preferred style of all stubs.

c) i have confirmed that this works with dropbox's server repo (having, now figured out that `update_stone` script exists.)

d) I assume the `bv`, `bb` verbiage is leftover from when Stone was Babel - should I change it?


braincore:
That's right. `bv`, `bb` are leftovers. `sv`, `sb` would be preferred now.

PR #74, opened 2018-02-26
python_type_stubs generator: use typing.Text instead of str

wittekm:
## What is typing.Text?
It's a type that means
if py2: means `unicode`
if py3: means `str` (which is a unicode object)

I was previously outputting just `str` which meant this behavior was wrong from py2 (in py2 world it expected non-unicode, ascii strings).

## Testing
There are unit tests done on Stone itself, and I confirmed this change is compatible with the current version of Dropbox's server repo after an `./update-stone` + `//tools:stone` + `mypy-run`

## Motivation for change
Existing behavior was a bug on my part.
Colin posted in Dropbox's api-platform slack that it was broken

### DIFF
--- a/stone/backends/python_type_mapping.py
+++ b/stone/backends/python_type_mapping.py
@@ -21,6 +21,7 @@ from stone.ir import (
     is_void_type,
 )
 from stone.backends.python_helpers import class_name_for_data_type
+from stone.ir.data_types import String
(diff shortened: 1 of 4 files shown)


wittekm:
(also, additionally, there are a few places throughout the Stone codebase itself - not just the code _generated by it_ - that also needed to be changed to typing.Text.

PR #76, opened 2018-03-03
Generate validators for all aliases, not just user-defined aliases.

wittekm:
last one guys, I swear.

Previously we weren't generating a validator in the PYI generator if I said
```
alias SomethingId = String(pattern="whatever")
```

This more closely follows the Python (non-pyi) generator.
https://github.com/dropbox/stone/blob/master/stone/backends/python_types.py#L155


scobbe:
Sorry for the delay!

wittekm:
No worries! I forgot to add a reviewer :)

PR #78, opened 2018-03-06
Add route whitelisting

JavaScriptBach:


stone/cli.py
```diff
@@ -104,6 +105,15 @@
           'combine multiple expressions with "and"/"or" and use parentheses '
           'to enforce precedence.'),
 )
+_cmdline_parser.add_argument(
+    '-r',
+    '--route-whitelist-filter',
+    type=six.text_type,
+    help=('Restrict datatype generation to only the routes specified in the whitelist '
+          'and their dependencies. Input should be a JSON file mapping namespace names to '
```
JavaScriptBach: oh, this doc should be updated to include `datatype_whitelist` as well.

stone/frontend/ir_generator.py
```diff
@@ -172,6 +181,8 @@ def generate_IR(self):
         self._populate_examples()
         self._validate_doc_refs()
         self._validate_annotations()
+        if self._routes is not None:
```
risforrob: this can be simplified to `if self._routes`.  The additional `is not None` is redundant.

stone/frontend/ir_generator.py
```diff
@@ -1248,3 +1259,235 @@ def mk_route_schema():
         # TODO: are we always guaranteed at least one data type?
         # pylint: disable=undefined-loop-variable
         return data_type
+
+    def _filter_namespaces_by_route_whitelist(self):
+        """
+        Given a parsed API in IR form, filter the user-defined datatypes
+        so that they include only the route datatypes and their direct dependencies.
+        """
+        assert self._routes is not None, "Missing route whitelist"
```
risforrob: Same as above, this can be simplified to `assert self._routes`

stone/frontend/ir_generator.py
```diff
@@ -1248,3 +1259,235 @@ def mk_route_schema():
...
+        assert 'route_whitelist' in self._routes
+        assert 'datatype_whitelist' in self._routes
+
+        # Parse the route whitelist and populate any starting data types
+        route_data_types = []
+        for namespace_name, routes_names in self._routes['route_whitelist'].items():
+            # Error out if user supplied nonexistent namespace
+            if namespace_name not in self.api.namespaces:
+                print('Namespace %s is not defined!' % namespace_name, file=sys.stderr)
+                sys.exit(1)
```
risforrob: Not noting enough about our stone build pipeline, is there a reason this is a print + sys.exit() instead of raising an exception that is handled elsewhere.  Calling sys.exit() this deep in a process does not seem appropriate.

stone/frontend/ir_generator.py
```diff
@@ -1248,3 +1259,235 @@ def mk_route_schema():
...
+            namespace = self.api.namespaces[namespace_name]
+
+            # Parse namespace doc refs and add them to the starting data types
+            if namespace.doc is not None:
```
risforrob: redundant `is not None`

stone/frontend/ir_generator.py
```diff
@@ -1248,3 +1259,235 @@ def mk_route_schema():
...
+            if routes_names == ['*']:
+                routes_names = namespace.route_by_name.keys()
+            #     route_data_types.extend(namespace.get_route_io_data_types())
+            # else:
+            assert '*' not in routes_names
+            for route_name in routes_names:
+                if route_name not in namespace.route_by_name:
+                    print('Route %s is not defined!' % route_name, file=sys.stderr)
+                    sys.exit(1)
+                route = namespace.route_by_name[route_name]
```
risforrob: A more "python-y" pattern for the logic in lines 1302-1305 would be as follows:

```python
route = namespace.route_by_name.get(route_name)
if not route:
    raise <exception>("message")
```

stone/frontend/ir_generator.py
```diff
@@ -1248,3 +1259,235 @@ def mk_route_schema():
...
+                route_data_types.extend(namespace.get_route_io_data_types_for_route(route))
+                if route.doc is not None:
```
risforrob: redundant `is not None`

stone/frontend/ir_generator.py
```diff
@@ -1248,3 +1259,235 @@ def mk_route_schema():
...
+            if namespace.doc is not None:
+                data_types, routes_by_ns = self._get_data_types_and_routes_from_doc_ref(
+                    namespace.doc, namespace.name)
+                for d in data_types:
+                    route_data_types.append(d)
+                for ns_name, routes in routes_by_ns.items():
+                    ns = self.api.namespaces[ns_name]
+                    for r in routes:
+                        for d in ns.get_route_io_data_types_for_route(r):
+                            route_data_types.append(d)
```
risforrob: This pattern in lines 1327 - 1335 shows up three different times in this method, is there a way to DRY this logic so you don't have the same code copied three times?

stone/frontend/ir_generator.py
```diff
@@ -1248,3 +1259,235 @@ def mk_route_schema():
...
+                data_type = self.api.namespaces[namespace_name].data_type_by_name[datatype_name]
+                route_data_types.append(data_type)
+
+        # Recurse on dependencies
+        output_types_by_ns, output_routes_by_ns = self._find_dependencies(route_data_types)
+
+        # Update the IR representation. This involves editing the data types and
+        # routes for each namespace.
+        for namespace in self.api.namespaces.values():
+            data_types = list(set(output_types_by_ns[namespace.name]))  # defaults to empty list
```
risforrob: FYI, technically sets() are also iterable.  Only reason you would need to convert a set() back to a list() is if you either need the resulting de-duplicated data accessible via index.

stone/frontend/ir_generator.py
```diff
@@ -1248,3 +1259,235 @@ def mk_route_schema():
...
+                        route_types = route_namespace.get_route_io_data_types_for_route(route)
+                        for route_type in route_types:
+                            self._find_dependencies_recursive(route_type, seen, output_types,
+                                                              output_routes)
+            if is_struct_type(data_type) and data_type.has_enumerated_subtypes():
+                for subtype in data_type.get_enumerated_subtypes():
+                    self._find_dependencies_recursive(subtype, seen, output_types, output_routes,
+                                                      type_context=data_type)
+        elif is_alias(data_type) or is_field_type(data_type):
+            assert (is_field_type(data_type)) == (type_context is not None)
```
risforrob: this is a strange way to write this assertion case.  Am I correct in understanding that you want `is_field_type(data_type) XOR (type_context)

stone/frontend/ir_generator.py
```diff
@@ -1248,3 +1259,235 @@ def mk_route_schema():
...
+            tag = match.group('tag')
+            val = match.group('val')
+            supplied_namespace = self.api.namespaces[namespace_context]
+            if tag == 'field':
+                if '.' in val:
+                    type_name, __ = val.split('.', 1)
+                    doc_type = supplied_namespace.data_type_by_name[type_name]
+                    data_types.add(doc_type)
+                else:
+                    pass  # no action required, because we must be referencing the same object
```
risforrob: else/pass is redundant

stone/cli.py
```diff
@@ -104,6 +105,15 @@
...
+          'either 1) a list of routes to generate, or 2) the character *, as shorthand '
+          'for generating all routes in that namespace.'),
+)
```
risforrob: Can we get either a comment or update the README with an example of what the whitelist file should look like.

stone/frontend/ir_generator.py
JavaScriptBach: yep, backend code (and mypy annotations) expect this to be a list, and uses some list-only functions too.

PR #85, opened 2018-04-12
Add versioning support to Python backends

starforever:
This is based on #82. Please ignore everything already included in #82.

### DIFF
--- a/stone/backends/python_helpers.py
+++ b/stone/backends/python_helpers.py
@@ -83,6 +83,21 @@ def fmt_version(version, prefix='_v'):
     else:
         return '{}{}'.format(prefix, version)
 
+def check_route_name_conflict(namespace):
+    """
+    Check name conflicts among generated route definitions. Raise a runtime exception when a
+    conflict is encountered.
+    """
+
+    route_by_name = {}
+    for route in namespace.routes:
+        route_name = '{}{}'.format(route.name, fmt_version(route.version))
+        if route_name in route_by_name:
+            other_route = route_by_name[route_name]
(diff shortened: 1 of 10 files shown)


stone/backends/python_types.py
```diff
@@ -877,7 +883,9 @@ def _generate_routes(self, route_schema, namespace):
         with self.block('ROUTES =', delim=('{', '}')):
             for route in namespace.routes:
                 var_name = fmt_func(route.name)
-                self.emit("'{}': {},".format(route.name, var_name))
+                self.emit('\'{}{}\': {}{},'.format(
```
qimingyuan: Why escape ' instead of use "" as outer quote?

stone/backends/python_helpers.py
```diff
@@ -77,6 +77,27 @@ def fmt_var(name, check_reserved=False):
...
+
+def check_route_name_conflict(namespace):
+    """
+    Check name conflicts among generated route definitions. Raise a runtime exception when a
+    conflict is encountered.
+    """
+
+    route_by_name = {}
+    for route in namespace.routes:
+        route_name = '{}{}'.format(route.name, fmt_version(route.version))
```
qimingyuan: Can we move `'{}{}'.format(route.name, fmt_version(route.version))` to some helper function so that we don't have to do so repeatedly?

starforever: I don't see much benefits of doing this. The only logic would be to take two strings and concatenate them together. It's too simple to worth a separate function.

PR #94, opened 2018-05-21
Update readme and test formatting

risforrob:
Added preferred style to readme and cleanup tab formatting in tests.

### DIFF
--- a/docs/lang_ref.rst
+++ b/docs/lang_ref.rst
@@ -315,21 +315,11 @@ Maps are expressed with curly braces::
 
 Map examples can also be multiline (Be mindful of indentation rules)::
 
-    struct SimpleDigits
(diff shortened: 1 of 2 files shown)


docs/lang_ref.rst
```diff
@@ -339,6 +330,18 @@ Map examples can also be multiline (Be mindful of indentation rules)::
                     "two": 22
                 }
             }
+    struct SimpleDigits
```
starforever: Would it be cleaner to just document the preferred style without explicitly stating it's preferred? This way users don't feel enforced to use a specific style (it's a preference after all). And by not advertising the other alternative we can implicitly boost the usage of the preferred one, but without making the users feel pushed to use it.

PR #97, opened 2018-06-22
Add versioning support to Swift SDK backend

starforever:


stone/backends/swift.py
```diff
@@ -142,6 +142,7 @@ def _docf(self, tag, val):
         if tag == 'route':
             if ':' in val:
                 val, version = val.split(':', 1)
+                version = int(version)
```
risforrob: Hah!  It's funny you made this change.  I actually thought about proposing this and then decided against it as I assumed you intentionally wanted to support non integer version numbers.

starforever: Haha, no, it has to be an integer as enforced by Stone syntax.

PR #98, opened 2018-06-22
Fix bugs in Python SDK backend for route versioning

starforever:
* Fixed route version in docstring.
* Code refactoring for fmt_func.

### DIFF
--- a/stone/backends/python_client.py
+++ b/stone/backends/python_client.py
@@ -150,8 +150,6 @@ class PythonClientBackend(CodeBackend):
         and routes are represented as Python classes."""
         self.cur_namespace = None
         for namespace in namespaces:
-            # Hack: needed for _docf()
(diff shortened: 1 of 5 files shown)


stone/backends/python_client.py
```diff
@@ -163,6 +161,9 @@ def _generate_routes(self, namespace):
         Generates Python methods that correspond to routes in the namespace.
         """
 
+        # Hack: needed for _docf()
+        self.cur_namespace = namespace
+
```
risforrob: Random curiosity, what's the reason for moving this into `_generate_routes()`?

starforever: Thanks for the quick approval! The tests call `_generate_routes` and `cur_namespace` needs to be set for `_docf` to work properly.

PR #99, opened 2018-06-27
Route versioning support for js and ts backends

pran1990:


pran1990:
https://github.com/dropbox/dropbox-sdk-js/pull/208 is how a change looks.

stone/backends/tsd_helpers.py
```diff
@@ -101,8 +101,18 @@ def fmt_union(type_strings):
...
 
-def fmt_func(name):
-    return fmt_camel(name)
+def fmt_func(name, version=1):
+    if version == 1:
+        return fmt_camel(name)
+    return fmt_camel(name) + 'V{}'.format(version)
+
+
+def fmt_url(namespace_name, route_name, route_version):
```
starforever: This function doesn't seem to be used?

stone/backends/tsd_helpers.py
```diff
@@ -101,8 +101,18 @@ def fmt_union(type_strings):
     """
     return '|'.join(type_strings) if len(type_strings) > 1 else type_strings[0]
 
-def fmt_func(name):
-    return fmt_camel(name)
+def fmt_func(name, version=1):
```
starforever: Code in https://github.com/dropbox/stone/pull/99/files#diff-48ab1e69649131b8fb683a0484b22bafR129 needs to be updated to handle route docstring properly. See https://github.com/dropbox/stone/pull/98 for reference.

Unless this function will be used in circumstances where version is not available, it might be better to make `version` a required field to make sure people won't forget to pass the parameter.

test/test_js_client.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class TestGeneratedJSClient(unittest.TestCase):
+    def __init__(self, *args, **kwargs):
+        # type: (ApiNamespace) -> typing.Text
+        super(TestGeneratedJSClient, self).__init__(*args, **kwargs)
+
+    def test_route_versions(self):
+        # type: () -> None
+        text = textwrap.dedent("""\
```
starforever: I'm thinking maybe we should consider constructing IR directly instead of from Stone using frontend. The unit tests for the backend should not have dependency on the frontend. This way if the tests fail we know there is something wrong on the backend.

stone/backends/tsd_helpers.py
pran1990: done. This was a good catch!  also added some versioned docstrings in tests to catch this.

pran1990:
https://github.com/dropbox/dropbox-sdk-js/pull/208 continues to reflect how a change looks.

stone/backends/js_client.py
```diff
@@ -104,5 +107,17 @@ def _docf(self, tag, val):  # pylint: disable=unused-argument
         Callback used as the handler argument to process_docs(). This converts
         Stone doc references to JSDoc-friendly annotations.
         """
-        # TODO(kelkabany): We're currently just dropping all doc ref tags.
+        # TODO(kelkabany): We're currently just dropping all doc ref tags ...
+        # NOTE(praneshp): ... except for versioned routes
```
starforever: Ha, I didn't expect you to fix this. Thanks!

stone/backends/js_client.py
```diff
@@ -104,5 +107,17 @@ def _docf(self, tag, val):  # pylint: disable=unused-argument
...
+        if tag == 'route':
+            if ':' in val:
+                val, version = val.split(':', 1)
+                version = int(version)
+            else:
+                version = 1
+            url = fmt_url(self.cur_namespace.name, val, version)
```
starforever: By my understanding the string returned is to reference a method not the URL. Should we use `fmt_func` instead?

stone/backends/tsd_client.py
```diff
@@ -102,17 +103,19 @@ def generate(self, api):
...
             for namespace in api.namespaces.values():
+                # first check for route name conflict
+                check_route_name_conflict(namespace)
                 for route in namespace.routes:
                     self._generate_route(
                         namespace, route)
 
     def _generate_route(self, namespace, route):
-        function_name = fmt_func(namespace.name + '_' + route.name)
+        function_name = fmt_func(namespace.name + '_' + route.name, version=route.version)
```
starforever: Mind removing `version=`?

test/test_js_client.py
```diff
@@ -0,0 +1,86 @@
...
+if MYPY:
+    import typing  # noqa: F401 # pylint: disable=import-error,unused-import,useless-suppression
+
+
+class TestGeneratedJSClient(unittest.TestCase):
+    def __init__(self, *args, **kwargs):
+        super(TestGeneratedJSClient, self).__init__(*args, **kwargs)
+
+    def _get_api(self):
+        # type () -> Api
```
starforever: Thanks for fixing the tests!

stone/backends/js_client.py
pran1990: Hmm. I am going to leave this one as a TODO. The reason is that the previous (empty) implementation of this function led to url type representations.

starforever: It's OK. I'm not sure but I think for SDK users, URLs should be transparent to them. They don't need to know the underlying details.

PR #109, opened 2018-08-14
support multiple auth modes in python SDK generation backend for stone

mghatak:
Added support for multiple auth modes in stone spec file in python backend code generation logic . This can now support comma separated auth modes in a stone spec file and puts the corresponding method declarations in either base.py or base_team.py.

Tested manually by updating the stone spec files for various combinations.


qimingyuan:
Actually this change breaks some tests.

PR #113, opened 2018-08-31
Fix Travis CI

starforever:


.travis.yml
```diff
@@ -1,27 +1,23 @@
...
 install:
     - pip install tox-travis
 
 matrix:
+    include:
+        # Workaround to enable Python 3.7: https://github.com/travis-ci/travis-ci/issues/9815
+        - python: 3.7
+          dist: xenial
+          sudo: true
     allow_failures:
```
pran1990: do we need this??

starforever: We used this to bypass failures on some configurations mostly due to temporary bugs. This fix removes all items currently here but I foresee we might need this in the future. It could be helpful to keep this stub to make it easier. I'll add some comment here.

starforever:
Thanks for the review! Will wait until EOD today to get this merged if no additional feedback is received.

PR #118, opened 2018-10-25
Fix aliases not being fully resolved by compiler.

aelawson:
Summary:

Code that handled `preserve_aliases=False` in `compiler.py` was not properly resolving the full chain (would stop with one more layer remaining).

I added a new `resolve_aliases` function that recursively resolves a chain of aliases to their concrete data type. I refactored the existing code to calls to this function.

I added a new test suite for the compiler module with cases that ensure 1) aliases are resolved to their underlying data type in the namespace when the `preserve_aliases` flag is set to `False` and 2) that they are present / point to the properly data type when it is set to `True`.

I added a new test suite for the data types module and added a case for the new `resolve_aliases` function.

Resolves `T158219`.


aelawson:
@starforever There's some lint errors on Travis for just the `py27` and `pypy` builds - I'm not sure why this is, it seems like only some warnings are thrown and no failures.

stone/backend.py
```diff
@@ -40,28 +41,14 @@ def remove_aliases_from_api(api):
...
-                data_type = field
-                while True:
-                    if hasattr(data_type, 'data_type'):
-                        if is_alias(data_type.data_type):
-                            data_type.data_type = data_type.data_type.data_type
-                        data_type = data_type.data_type
-                    else:
-                        break
+                if hasattr(field, 'data_type'):
+                    field.data_type = resolve_aliases(field.data_type)
```
starforever: Since we already have all aliases resolved in above code, maybe it's easier to just write (knowing there is no intermediate alias):
```
if is_alias(field.data_type):
    field.data_type = field.data_type.data_type
```

This aligns with how route args are handled below.

test/test_compiler.py
```diff
@@ -0,0 +1,160 @@
...
+    PythonTypesBackend
+)
+from stone.compiler import (
+    Compiler
+)
+from stone.frontend.frontend import (
+    specs_to_ir
+)
+
+class PreserveBackend(PythonTypesBackend):
```
starforever: Can we inherit from `CodeBackend` to make the tests language-independent?

test/test_data_types.py
```diff
@@ -0,0 +1,39 @@
...
+from stone.ir import (
+    resolve_aliases,
+    Alias,
+    String
+)
+
+class TestDataTypes(unittest.TestCase):
+    """Tests for the data_types module"""
+
+    def test_resolve_aliases(self):
```
starforever: Would it make sense to move this test to `test_backend` as well? `resolve_aliases` should only be called by the backend.

starforever:
The lint is only enabled on `py27` and `pypy` for current settings. That's why you only see errors there. Most of the warnings are W504 which seems a recent bug. I'm going to send a PR to fix this. You can ignore them for now. There are a few other lint errors in the new files which should be addressed. I think in current settings we treat warnings as errors to ensure good code quality.

stone/backend.py
```diff
@@ -40,28 +41,14 @@ def remove_aliases_from_api(api):
...
-                        # Skip the alias (this looks so terrible...)
-                        data_type.data_type = data_type.data_type.data_type
-                    data_type = data_type.data_type
-                else:
-                    break
-
+            # For better or for worse, all non-user-defined types that
+            # reference other types do so with a 'data_type' attribute.
+            if hasattr(alias, 'data_type'):
+                resolve_aliases(alias.data_type)
```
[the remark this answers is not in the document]
aelawson: It just needs to be `resolve_aliases(alias.data_type)`. Although looking at this again, I think it would be much better to pass in the parent `alias` obj like in your example (don't remember what my reasoning was for passing in `alias.data_type`).

test/test_data_types.py
aelawson: That sounds fine - I just created this module since the function belongs to the `data_types` module and I anticipate there will be more tests added there in the future. Whichever you think is best!

test/test_compiler.py
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
aelawson: Okay - I leaned on doing a complete end-to-end test here over just unit tests, but that makes sense / is fine by me.

test/test_compiler.py
aelawson: Currently the test cases using these classes are expecting a Python module to be used with `pytest` (so there's some coupling to a language here). I think changing the end-to-end test mentioned above to modular unit tests will remove the need for these / this module anyways.

aelawson:
Okay - I'll squash the commits for this one after review is completed and next time will open a PR from a fork.

I'll take a look at the lint job again - I merged in the bug fix for the `W504`.

test/test_data_types.py
starforever: Yeah, it's generally reasonable to have test after each file. However it's not how existing tests are organized. And I feel it's kind of confusing to have mixed organization.

test/test_compiler.py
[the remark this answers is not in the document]
starforever: > Okay - I'll squash the commits for this one after review is completed and next time will open a PR from a fork.
> 
> I'll take a look at the lint job again - I merged in the bug fix for the `W504`.

👍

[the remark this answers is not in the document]
starforever: Ideally we should have both end-to-end and unit tests. I just feel end-to-end tests are less informative than unit tests when identifying the issue. So I prefer unit tests over end-to-end tests if I only have resource to write one.

disclaimer: not a professional tester.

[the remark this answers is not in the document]
aelawson: Haha yeah makes sense 👍

stone/ir/data_types.py
```diff
@@ -1933,6 +1933,32 @@ def unwrap_aliases(data_type):
...
+    def _resolve_data_type_chain(data_type):
+        if not is_alias(data_type):
+            return data_type
+
+        resolved = _resolve_data_type_chain(data_type.data_type)
+        data_type.data_type = resolved
+
+        return resolved
+
+    if hasattr(alias, 'data_type'):
```
starforever: Hmm. I might miss something here. But what's the point of this checking? If `alias` is of type `Alias`, then `data_type` should always present since it's set in `__init__` (https://github.com/dropbox/stone/blob/6e8744a735eb9b68bce5730f5f2ea80cca5d5081/stone/ir/data_types.py#L1816).

I also don't see the point of wrapping this in an internal function. The original implementation looks good to me.

aelawson: Yeah I made a note to remove that check :)

I changed the implementation since it seemed a bit weird/arbitrary to expect a developer to pass in the alias's `data_type` attribute e.g. `resolve_aliases(alias.data_type)` as opposed to just passing in the alias e.g. `resolve_aliases(alias)` and handling it internally.

aelawson:
Oh weird... That's what I thought I did, but I see that merge got a bit messed up (not sure how). I'll fix it.

aelawson:
@starforever I've finished making the changes we talked about - let me know if there's anything else you have questions/concerns about.

I'm anticipating squashing all these commits as part of the merge process (via Squash and Merge).

stone/ir/data_types.py
```diff
@@ -1933,6 +1933,31 @@ def unwrap_aliases(data_type):
...
+            return data_type
+
+        resolved = _resolve_data_type_chain(data_type.data_type)
+        data_type.data_type = resolved
+
+        return resolved
+
+    alias.data_type = _resolve_data_type_chain(alias.data_type)
+
+    return alias
```
starforever: I'm thinking if we just return whatever alias passed in, it's not something quite useful. Should we simply return the resolved data type (as stated in the comment on line 1947)?. If we change it this way, we could avoid the internal function `_resolve_data_type_chain` to make the code cleaner.

test/test_backend.py
```diff
@@ -225,6 +233,154 @@ def test_backend_cmdline(self):
...
+alias StructAlias = TestStruct
+
+struct TestResult
+    field1 PathRootId
+
+union TestError
+    test PathRootId
+"""
+
+        api = specs_to_ir(
```
starforever: Can we construct IR directly instead of from the Stone code, like how it's done in `test_api_namespace`? This removes the dependency on frontend code so that the tests won't fail due to bugs there.

stone/ir/data_types.py
aelawson: Okay I'll just revert to my original implementation.

risforrob:
🎉

PR #119, opened 2018-10-27
XCode 10 shows a warning against use of 'open let'

flexfrank:
This pull request fixes the issue https://github.com/dropbox/SwiftyDropbox/issues/228


greg-db:
Thanks for putting this together!

PR #128, opened 2019-01-24
Move typing imports to top of pyi files so they'll get used

lazytype:
So I discovered that because we're adding the imports from the `typing` module at the bottom of the file, it's not getting picked up by mypy. I was able to update the code to move the imports to the top in this diff.
 
However, if we try to import this into the DB repo we'll get a bunch of type errors that need to be addressed. Most of them fall under one of two classes:
1. need to assert an `Optional` value is not `None`
2. need to handle `unicode` strings. (Because before, the `Text` type was just being treated as `Any`. Now that it's working properly, `Text` will indicate when something can be `str` or `unicode` and a lot of code only handles the `str` case.)

I think I can handle fixing all cases of (1) by adding the necessary type checks.
I think there are just way too many cases of (2). My short term solution (included in this diff) is to just export `Any` in place of `Text` (which would be consistent with the existing behavior) and then change it back after we fix all cases of (2).


lazytype:
Duplicate of #134

PR #134, opened 2019-03-15
Move typing imports to top of pyi files so they'll get used

### LINKED ISSUE #107: `python_type_stubs` generator incorrectly puts imports at bottom of file, not top
Hi all! This is actually a bug I introduced, just making sure it's tracked somewhere.

# The issue
Mypy *will not* correctly detect the type of `a` in the following code snippet of a `.pyi` file.
```
a = ...  # type: Text
from typing import Text
```
Mypy *will* correctly detect the type of `a` if we literally just swap those lines.
```
from typing import Text
a = ...  # type: Text
```

Unfortunately, my `python_type_stubs` generator does exactly the incorrect behavior, as shown in this test suite:
https://github.com/dropbox/stone/blob/4f4e5906a2be1431ea6f12726959f456d0670086/test/test_python_type_stubs.py#L306

The result: Our apiv2 stuff is often poorly typed as `typing.Any`s :(

# Where does the issue occur?
If we look at  [_generate_base_namespace_module](https://github.com/dropbox/stone/blob/4f4e5906a2be1431ea6f12726959f456d0670086/stone/backends/python_type_stubs.py#L113), sure enough the `_generate_imports_needed_for_typing` is the final thing called. This is because we build up a list of all the things that, well, _do_ need to be imported.

# The solution
Just my opinion, feel free to come up with another solution:

`self.emit_raw` should not immediately write to a file - perhaps split this into something like `self.append_to_output_buffer(...)`, that doesn't actually write anything to a file yet, and a `self.emit_output_buffer(...)`. This would let us do something like:

```
<stone doing its' thing, doing self.append_output_buffer() while figuring out what imports are needed>
self.prepend_output_buffer(the imports required)
self.emit_output_buffer
```

ramosbugs:
This PR is a port of #128 rebased onto latest Stone master,
with the Any types changed back to Text.

Resolves #107.


ramosbugs:
cc: @lazytype

lazytype:
Going to close #128 then if that's alright?

pran1990:
thanks ramos!

keep a watch out for anything random!

wittekm:
thanks!!

PR #135, opened 2019-03-26
Add field path to custom annotations processing

ramosbugs:
This diff adds a `field_path` as the second argument to the custom
annotations processor. Custom processors can use this string to
provide helpful error messages that identify the field in which an
error occurs.

### DIFF
--- a/stone/backends/python_rsrc/stone_base.py
+++ b/stone/backends/python_rsrc/stone_base.py
@@ -31,8 +31,8 @@ if _MYPY:
 
 class Struct(object):
     # This is a base class for all classes representing Stone structs.
-    def _process_custom_annotations(self, annotation_type, processor):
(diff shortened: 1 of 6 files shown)


stone/backends/python_rsrc/stone_base.py
```diff
@@ -73,8 +73,8 @@ def __ne__(self, other):
     def __hash__(self):
         return hash((self._tag, self._value))
 
-    def _process_custom_annotations(self, annotation_type, processor):
-        # type: (typing.Type[T], typing.Callable[[T, U], U]) -> None
+    def _process_custom_annotations(self, annotation_type, context, processor):
```
jiuyangzhao: Since you are going to change context in `make_list_annotation_processor` and `make_map_value_annotation_processor`. I suggest we make it clear the the second parameter is the path.

PR #136, opened 2019-03-30
Use six to set Validator metaclass to ABCMeta

gvanrossum:
This file uses six so clearly it's meant to work with Python 3, but
the way it was written it wouldn't set the metaclass in that case.


yuxiang-he:
Actually #154 has made the same change

PR #137, opened 2019-05-16
Use better relative imports for dummy namespace modules

ilevkivskyi:
This will cause generating the following content for e.g. `.../async.py`:
```python
# If you have issues importing this module because Python recognizes it as a keyword, use async_ instead.
from .async_ import *
```
Instead of
```python
# If you have issues importing this module because Python recognizes it as a keyword, use async_ instead.
from async_ import *
```
The former can cause troubles with mypy.


ilevkivskyi:
I am trying to investigate the test failures in CI. Everything passes locally on my machine on both Python 3.6 and 3.7. Does anyone have any hints?

greg-db:
@ilevkivskyi Can you share some more information about the issue? Is there any error output you can share?

ilevkivskyi:
> Can you share some more information about the issue? Is there any error output you can share?

It is right here in Travis CI for this PR https://travis-ci.org/dropbox/stone/builds/533346407?utm_source=github_status&utm_medium=notification

Some failures look like an infra failure (you need to update Python version in requirements to run mypy). But there are also test failures I can't reproduce.

greg-db:
Thanks! We'll take a look.

ilevkivskyi:
I already pushed a CI update here.hopefully it will at least fix the mypy failures.

ilevkivskyi:
OK, now the failures are only on Python 2 and PyPy 2.

ilevkivskyi:
OK, all is green now.

PR #138, opened 2019-05-22
Check if data_type is map and format as key value pair

adrian2x:
`Map(String, T)` types will be converted to a typescript key value pair like `{[key: string]: T}`, instead of plain `Object`.


greg-db:
Thanks!

PR #139, opened 2019-06-14
update compatibility doc - remove old union catch-all references

mghatak:


docs/evolve_spec.rst
```diff
@@ -49,7 +49,7 @@ Backwards Incompatible Changes
   * An old receiver may have application-layer dependencies on the field
     type. In statically typed languages deserialization will fail.
 
-* Adding a new tag to a union without a `catch-all tag <lang_ref.rst#union-catch-all>`_.
+* Adding a new tag to a closed union.
```
greg-db: Nit: Let's be consistent about "a new tag" and "another tag" between here and the next part. I think "a new tag" is slightly better.

docs/evolve_spec.rst
```diff
@@ -90,19 +90,13 @@ Backwards Compatible Changes
...
 
-* Adding another tag to a union that has a catch-all specified.
+* Adding another tag to an open union.
 
   * The older receiver will not understand the incoming tag, and will
     simply set the union to its catch-all tag. The application-layer will
     handle this new tag through the same code path that handles the
     catch-all tag.
 
-Planning for Backwards Compatibility
```
greg-db: I know the default is an open `union` now, but I think it's still worth calling out the difference, and how the developer should choose between the two. So, let's keep this section but explain when one should use open or closed.

docs/evolve_spec.rst
```diff
@@ -90,19 +90,13 @@ Backwards Compatible Changes
     data type and continue to present a tag with no value to the
     application.
 
-* Adding another tag to a union that has a catch-all specified.
+* Adding another tag to an open union.
```
greg-db: Per the above, let's switch this to "a new tag" to match.

PR #141, opened 2019-08-12
Proto gen data types

Mr-Gupta:
Adds basic data type mappings


stone/backends/proto_gen/proto_type_mapping.py
```diff
@@ -0,0 +1,60 @@
...
+        return 'uint32'
+
+    elif is_uint64_type(data_type):
+         return 'uint64'
+
+    elif is_float32_type(data_type):
+        return 'float32'
+
+    elif is_float64_type(data_type):
+        return 'float64'
```
aelawson: Shouldn't this be `double`?

PR #150, opened 2020-01-16
turn off CI for Python 3.4

yuxiang-he:


yuxiang-he:
Python 3.4 is failing tests because of exit code 1 of tox. The error does not seem to be related to any particular commit and Python 3.4 has reached the end of lifecycle. Ignoring Python 3.4 failures for now

PR #152, opened 2020-01-18
Make Mypy annotations Python Stub files more sensible

yuxiang-he:
Previously the constructor in stub files has `Optional` for all fields. This PR makes only nullable fields and fields with default values `Optional`.

### DIFF
--- a/stone/backends/python_type_stubs.py
+++ b/stone/backends/python_type_stubs.py
@@ -334,7 +334,7 @@ class PythonTypeStubsBackend(CodeBackend):
             field_name_reserved_check = fmt_var(field.name, True)
             field_type = self.map_stone_type_to_pep484_type(ns, field.data_type)
 
-            if not is_nullable_type(field.data_type):
+            if field.has_default:
                 self.import_tracker._register_typing_import('Optional')
                 field_type = 'Optional[{}]'.format(field_type)
 
--- a/test/test_python_type_stubs.py
+++ b/test/test_python_type_stubs.py
@@ -210,6 +210,42 @@ def _make_namespace_with_route_name_conflict():
     ns.add_route(route_two)
     return ns
 
+def _make_namespace_with_nullable_and_dafault_fields():
+    # type: (...) -> ApiNamespace
+    ns = ApiNamespace('ns_w_nullable__fields')
+
+    struct = Struct(name='Struct1', namespace=ns, ast_node=None)
+    default_field = StructField(
+                        name='DefaultField',
+                        data_type=UInt64(),
+                        doc=None,
+                        ast_node=None,
+                    )
+    default_field.set_default(1)
+    struct.set_attributes(
+        doc=None,
+        fields=[
+            StructField(
+                name='NullableField',
+                data_type=Nullable(
+                    UInt64()
+                ),
+                doc=None,
+                ast_node=None,
+            ),
+            default_field,
+            StructField(
+                name='NonNullableField',
+                data_type=UInt64(),
+                doc=None,
+                ast_node=None,
+            )
+        ]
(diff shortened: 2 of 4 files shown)


stone/backends/python_type_stubs.py
```diff
@@ -334,7 +334,7 @@ def _generate_struct_class_init(self, ns, struct):
             field_name_reserved_check = fmt_var(field.name, True)
             field_type = self.map_stone_type_to_pep484_type(ns, field.data_type)
 
-            if not is_nullable_type(field.data_type):
```
KarandeepJohar: Why was this `not is_nullable_type` in the first place? Was that just a mistake?

stone/backends/python_type_stubs.py
```diff
@@ -334,7 +334,7 @@ def _generate_struct_class_init(self, ns, struct):
...
+            if field.has_default:
```
KarandeepJohar: Do Nullable types have `has_default` as true? 

Why not just make it `if is_nullable_type(field.data_type)`

test/test_python_type_stubs.py
```diff
@@ -341,7 +376,7 @@ def test__generate_base_namespace_module__with_nested_types(self):
 
             class NestedTypes(bb.Struct):
                 def __init__(self,
-                             list_of_nullables: Optional[List[Optional[int]]] = ...,
+                             list_of_nullables: List[Optional[int]] = ...,
```
KarandeepJohar: Looking at the corresponding code
`            StructField(
                name='NullableList',
                data_type=Nullable(
                    List(UInt64())
                ),
                doc=None,
                ast_node=None,
            ),`

shouldn't this be `List[int]` instead?

test/test_python_type_stubs.py
```diff
@@ -210,6 +210,42 @@ def _make_namespace_with_route_name_conflict():
...
+                name='NullableField',
+                data_type=Nullable(
+                    UInt64()
+                ),
+                doc=None,
+                ast_node=None,
+            ),
+            default_field,
+            StructField(
+                name='NonNullableField',
```
KarandeepJohar: Nit: just call this required field?

stone/backends/python_type_stubs.py
yuxiang-he: Yeah this is the line that has a lot of history behind. The purpose of this PR is basically roll back the change in this line. T329658 keeps track of it. The long story short is that #126 and #134  introduced changes that basically makes every constructor arg and getter return type optional. This is not ideal, and #131 removed the issue with the getters, but did not touch the issue with the constructor. This PR changes that by only allowing values with default and nullable values to be `Optional`.

stone/backends/python_type_stubs.py
yuxiang-he: Yes Nullable types have default values https://github.com/dropbox/stone/blob/7aad3c33cb890bd728ddd21976c867262663d943/stone/backends/python_rsrc/stone_validators.py#L618-L622

There is another case beside Nullable types. Types with default values that are not `None` should also be optional. `field.has_default` captures both cases

test/test_python_type_stubs.py
yuxiang-he: Alright will change

test/test_python_type_stubs.py
yuxiang-he: Actually the definition of this arg `list_of_nullables` is
```
StructField(
                name='ListOfNullables',
                data_type=List(
                    Nullable(UInt64())
                ),
                doc=None,
                ast_node=None,
            )
```
so I think it should be `List[Optional[int]]` here

PR #155, opened 2020-02-02
Ensure equality checks work for Stone structs and unions

lazytype:
This does two things:
1) Implements a non-hacky equality function for Stone structs
2) Makes the typing of Stone union tags consistent. Currently, in Python 2, code generated union tags are of type `bytes`, but deserialized union tags are of type `unicode`. This causes defining equality through `repr` to not work, since `'u"tag_name"'` will not equal `'"tag_name"'`. Ensuring that the tag is `str` will guarantee the correct behavior. (In Python 2, `str` is `bytes` and in Python 3, `str` is `unicode`).

### DIFF
--- a/test/test_python_gen.py
+++ b/test/test_python_gen.py
@@ -48,7 +48,7 @@ class TestDropInModules(unittest.TestCase):
         # Passes
         s.validate('a')
         # Check that the validator is converting all strings to unicode
-        self.assertEqual(type(s.validate('a')), six.text_type)
+        self.assertIsInstance(s.validate('a'), six.text_type)
 
     def test_string_regex_anchoring(self):
         p, f = self.mk_validator_testers(bv.String(pattern=r'abc|xyz'))
@@ -527,10 +527,13 @@ class TestDropInModules(unittest.TestCase):
         # Test void variant
         u = json_decode(bv.Union(U), json.dumps('b'))
         self.assertEqual(u._tag, 'b')
+        self.assertIsInstance(u._tag, str)
+
         self.assertRaises(bv.ValidationError,
                           lambda: json_decode(bv.Union(U), json.dumps({'b': [1, 2]})))
         u = json_decode(bv.Union(U), json.dumps({'b': [1, 2]}), strict=False, old_style=True)
         self.assertEqual(u._tag, 'b')
+        self.assertIsInstance(u._tag, str)
 
         # Test struct variant
         u = json_decode(bv.Union(U), json.dumps({'c': {'f': 'hello'}}), old_style=True)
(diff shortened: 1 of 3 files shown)


test/test_python_gen.py
```diff
@@ -1198,6 +1217,55 @@ def test_struct_decoding_with_optional_struct(self):
...
+        # S_extend and S_extend2 are indirectly related because they both extend S, but they do not
+        # have a direct line of inheritance to each other.
+        s_extend = self.decode(self.sv.Struct(self.ns.SExtend), json.dumps({'f': 'F', 'g': 'G'}))
+        s_extend2 = self.decode(self.sv.Struct(self.ns.SExtend2), json.dumps({'f': 'F', 'g': 'G'}))
+        self.assertNotEqual(s_extend, s_extend2)
+        # They still hash to the same value, since they have the same fields, but this is
+        # fine since we don't expect to use a large number of structs as dict keys.
+        self.assertEqual(hash(s_extend), hash(s_extend2))
+
+    def test_extended_struct_equality(self):
```
yuxiang-he: I'm not sure if we should treat a subtype and its super type with the same value in their common fields to be equal, after all a subtype has extra fields that distinguish itself from its super type.

test/test_python_gen.py
```diff
@@ -1198,6 +1217,55 @@ def test_struct_decoding_with_optional_struct(self):
...
+        """Structs which subclass each other are considered equal to each other."""
+        s = self.decode(self.sv.Struct(self.ns.S), json.dumps({'f': 'F'}))
+        s_extend = self.decode(self.sv.Struct(self.ns.SExtend), json.dumps({'f': 'F', 'g': 'G'}))
+        s_extend_extend = self.decode(self.sv.Struct(self.ns.SExtendExtend),
+                                      json.dumps({'f': 'F', 'g': 'G', 'h': 'H'}))
+        self.assertEqual(s, s_extend)
+        self.assertNotEqual(hash(s), hash(s_extend))
```
yuxiang-he: Two points on `__eq__` and `__hash__`
1. The hash of an object does not change across the object's lifetime
2.  a == b must imply hash(a) == hash(b). The reverse is not necessarily true

From the Python Docs [here](https://docs.python.org/3.7/reference/datamodel.html#object.__hash__)
>  If a class defines mutable objects and implements an `__eq__()` method, it should not implement `__hash__()`, since the implementation of hashable collections requires that a key’s hash value is immutable (if the object’s hash value changes, it will be in the wrong hash bucket).

Although the Python Docs do not call out having `True` for `__eq__` implies same hash, the default behavior for user defined class shows Python follows this invariant
>User-defined classes have `__eq__()` and `__hash__()` methods by default; with them, all objects compare unequal (except with themselves) and `x.__hash__()` returns an appropriate value such that `x == y` implies both that x is y and `hash(x) == hash(y)`.

See the articles [here](https://www.asmeurer.com/blog/posts/what-happens-when-you-mess-with-hashing-in-python/) and [here](https://hynek.me/articles/hashes-and-equality/) for detailed examples of the two points

PR #160, opened 2020-05-21
change objc description method to debugDescription

zhaoze1991:
the objc generated API object code looks like:
```
- (NSString *)description {
  return [[XYZSerializer serialize:self] description];
}
```
The API object returned from server is very likely contains PII.
If we unintentionally logged API object, we would leak PII.

This diff changed description method to debugDescription so we won't leak PII
even we logged the API object unintentionally.


jboulter11:
This looks fine to me, but I'd feel more comfortable if you got a second opinion from someone who knows this codebase, and the repercussions of this change, better.

bogo:
`NSObject` contains a default implementation of `-description`, which should be sufficient. 👍

zhaoze1991:
thanks @jboulter11 and  @bogo

PR #162, opened 2020-06-13
Drop tests for Python 2.x and 3.4

connorworley:
These versions are EOL and the tests have been broken against them for a while. Let's just stop testing against them.


.travis.yml
```diff
@@ -1,27 +1,13 @@
 language: python
 
 python:
-    - 2.7
-    - 3.4
     - 3.5
     - 3.6
-    - pypy
+    - 3.7
```
aelawson, line 6: Should we also add 3.8?

connorworley: I don't think it's packaged on xenial or bionic, but maybe focal. Let me see what we can do with travis.

aelawson: Cool if it's non-trivial we don't have to block this PR on it. Would be nice to have though.

connorworley: Looks like some tests fail: https://travis-ci.org/github/dropbox/stone/jobs/697871257

PR #164, opened 2020-07-08
Stop generating relative paths for Python

connorworley:
This is a breaking change to simplify the output of the python backends. The current behaviour for imports in generated files is to first try a relative import, then try an absolute import that only works if the parent directory is a pythonpath. Neither of these really mesh well with our internal bazel stuff, and the value of keeping the relative imports seems pretty dubious to me, so I've refactored the backends to only generate absolute imports.

Two things are backwards incompatible:
* The `python_types` and `python_type_stubs` backends now require a package name as an argument in order to form absolute imports.
* The static files in `stone/backends/python_rsrc/` are no longer copied into the output directory. Instead they're imported from `stone` itself, meaning the output now depends on stone.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ with open('README.rst') as f:
 
 dist = setup(
     name='stone',
-    version='1.3.2',
+    version='2.0.0',
     install_requires=install_reqs,
     setup_requires=setup_requires,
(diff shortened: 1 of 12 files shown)


connorworley:
I'm going to go ahead and land this as the internal equivalent has already landed.

PR #167, opened 2020-07-16
Use __slots__ for validator classes

ilevkivskyi:
This allows to save some memory for large code-bases with many attributes. This is technically not 100% backwards compatible (because class variables can't match slot names), however I run our internal tests and only one test failed (out of many thousands), so I assume this is OK.

### DIFF
--- a/stone/backends/python_rsrc/stone_validators.py
+++ b/stone/backends/python_rsrc/stone_validators.py
@@ -92,6 +92,7 @@ def generic_type_name(v):
 
 class Validator(six.with_metaclass(ABCMeta, object)):
     """All primitive and composite data types should be a subclass of this."""
+    __slots__ = ("_redact",)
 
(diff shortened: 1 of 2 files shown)


stone/backends/python_rsrc/stone_validators.py
```diff
@@ -125,8 +128,10 @@ class Integer(Primitive):
     Do not use this class directly. Extend it and specify a 'minimum' and
     'maximum' value as class variables for a more restrictive integer range.
```
JukkaL, line 129: The docstring is out of date.

ilevkivskyi: Good catch!

ilevkivskyi:
I am assuming there are no objections, so I am going to merge this now (to move forward with using this internally).

PR #170, opened 2020-07-26
Modernize for Python 3!

connorworley:
This is a big mostly-automated change to modernize Stone's code.

[pyupgrade](https://github.com/asottile/pyupgrade) (targeting py36+)
* Ran on everything except `stone/backends/python_rsrc/` and `setup.py`

[com2ann](https://github.com/ilevkivskyi/com2ann)
* Ran on everything except `stone/backends/python_rsrc/` and `setup.py`

Black (targeting py36)
* Ran on everything except `stone/backends/python_rsrc/`

Black (targeting py27)
* Ran on `stone/backends/python_rsrc/` and `setup.py`

isort
* Ran on everything

Note that the files in `stone/backends/python_rsrc/` need to remain Py2 compatible for internal use. It might make sense to move them to a separate package entirely to avoid special-casing them.

I also simplified `setup.py` and `tox.ini`, and added py38 to the test suite.

I'll test internally before merging.


connorworley:
I realized I'm going to have to split out the python_rsrc stuff to another package before landing this if I want to vendor it in a sane way. Going to hold off on merging for now.

PR #178, opened 2020-09-18
Include module with type name in validation errors (python)

connorworley:
This is helpful for debugging when types that share the same name but live in different namespaces are mistakenly used with each other.


stone/backends/python_rsrc/stone_validators.py
```diff
@@ -69,6 +69,11 @@ def __repr__(self):
         return 'ValidationError(%r)' % six.text_type(self)
 
 
+def type_name_with_module(t):
+    # type: (typing.Type[typing.Any]) -> six.text_type
+    return '%s.%s' % (t.__module__, t.__name__)
+
+
 def generic_type_name(v):
     """Return a descriptive type name that isn't Python specific. For example,
```
KarandeepJohar, line 78: can you add types here as well?

stone/backends/python_rsrc/stone_validators.py
```diff
@@ -69,6 +69,11 @@ def __repr__(self):
         return 'ValidationError(%r)' % six.text_type(self)
 
 
+def type_name_with_module(t):
```
KarandeepJohar, line 72: shouldn't we always use `generic_type_name` instead of this right?

KarandeepJohar (changes requested):

stone/backends/python_rsrc/stone_validators.py
connorworley: I don't think that `generic_type_name` will work for the `self.definition` case later on - it'll end up returning `builtins.type`.

rogebrd (approved):
LGTM

PR #186, opened 2020-10-05
Update JS and TS routes for null parameters

rogebrd:
- Fix for ([JS #41](https://github.com/dropbox/dropbox-sdk-js/issues/41))
- Update route to not have `arg` when there is no parameter
- Update comments to also not contain `arg` when no parameter
- Update tests for both clients

## **Checklist**

**General Contributing**
- [X] Have you read the Code of Conduct and signed the [CLA](https://opensource.dropbox.com/cla/)?

**Is This a Code Change?**
- [X] Code Change
- [X] Example/Test Code Change

**Validation**
- [X] Have you ran `tox`?
- [X] Do the tests pass?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ with open('README.rst') as f:
 
 dist = setup(
     name='stone',
-    version='2.2.2',
+    version='3.0.0',
     install_requires=install_reqs,
     setup_requires=setup_requires,
     tests_require=test_reqs,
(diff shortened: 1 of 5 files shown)


connorworley (approved):

aelawson (changes requested):
Looks good, but we need to update the version (this is also a breaking change so should should be major version bump).

aelawson (approved):

PR #190, opened 2020-10-09
Fix doc string in argument only shows nullable without wrapping data …

jiuyangzhao:
…type.

Fixing "The Dropbox API v2 Python SDK docs don't show the parameter type when the parameter is optional"

## **Checklist**

**General Contributing**

**Is This a Code Change?**
- [x] Code Change

**Validation**
- [x] Have you ran `tox`?
- [ ] Do the tests pass
There are irrelevant test failure (all in JS/TS). I assume it was because the base commit is old. I didn't use the latest commit because of the -p issue.

### DIFF
--- a/stone/backends/python_client.py
+++ b/stone/backends/python_client.py
@@ -18,6 +18,8 @@ from stone.backends.python_types import (
 )
 from stone.ir import (
     is_nullable_type,
+    is_list_type,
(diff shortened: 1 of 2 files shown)


rogebrd (approved):
Looks good but please make sure CI passes before merging

PR #192, opened 2020-10-14
Add optional flag allowing wrapping responses in a parent class for J…

rogebrd:
…S and TS client

- Adds the `--wrap-response-in` flag which allows users to pass in a value
- Response will be wrapped in the following way: `wrapper<type>` where `wrapper` is the string passed in with the flag, and `type` is the original response type

## **Checklist**

**General Contributing**
- [X] Have you read the Code of Conduct and signed the [CLA](https://opensource.dropbox.com/cla/)?

**Is This a Code Change?**
- [X] Code Change

**Validation**
- [X] Have you ran `tox`?
- [X] Do the tests pass?


stone/backends/js_client.py
```diff
@@ -81,11 +87,21 @@ def _generate_route(self, route_schema, namespace, route):
         if route.deprecated:
             self.emit(' * @deprecated')
 
+        return_type = None
+        if self.args.wrap_response_in:
+            if route.result_data_type.__class__ != Void:
```
aelawson, line 92: Isn't there value in still doing this for Void type? e.g. if you are providing access to headers with a Response wrapper class like we do in the SDK?

stone/backends/tsd_client.py
```diff
@@ -126,13 +132,21 @@ def _generate_route(self, namespace, route):
             self.emit(' * @param arg The request parameters.')
         self.emit(' */')
 
-        if route.arg_data_type.__class__ != Void:
-            self.emit('public %s(arg: %s): Promise<%s>;' %
-                    (function_name, fmt_type(route.arg_data_type),
-                    fmt_type(route.result_data_type)))
+        return_type = None
+        if self.args.wrap_response_in:
+            if route.result_data_type.__class__ != Void:
```
aelawson, line 137: Same as above.

aelawson (approved):
This seems fine to me

PR #201, opened 2020-11-09
Fix obj_c types to not overwrite `jsonDict` with nil

geoffmacd:
When the inner subType serializer returns nil as it does with `return [jsonDict count] > 0 ? jsonDict : nil`, this causes the outer/parent serializer to overwrite the mutable `jsonDict` to nil, resulting in a nil result when the `.tag` field was going to be added... In our case the  `DBCAMERAUPLOADSMOBILEMediaMetadataSerializer` serializer returned nil when it needed a `.tag` field to be successful server-side

**General Contributing**
- [x ] Have you read the Code of Conduct and signed the [CLA](https://opensource.dropbox.com/cla/)?

**Is This a Code Change?**
- [ x] Code Change

**Validation**
- [x] Have you ran `tox`?
- [x] Do the tests pass?


stone/backends/obj_c_types.py
```diff
@@ -1120,7 +1120,7 @@ def emit_serializer():
                         if is_user_defined_type(data_type):
                             if is_struct_type(data_type) and \
                                     not data_type.has_enumerated_subtypes():
-                                self.emit('jsonDict = [{} mutableCopy];'.
```
rafe-g, line 1123: The only thing i'd be worried about here is that in the old version, we are relying on the fact that we either get a brand new dictionary or nil. I the new code, we are adding the serialized dictionary to whatever `jsonDict` already contains. Are we 100% certain that every instance where this is used does not have an instance of `jsonDict` with some values already in it?? 

Something safer (if your unsure about my question above) would be to do something like:

```
jsonDict = [{} mutableCopy] ?: [[NSMutableDictionary alloc] init];
```

geoffmacd: this is the deleted code, I think your saying that I should do 
```
let additionDict = [{} mutableCopy] ?: [[NSMutableDictionary alloc] init]; // to guarantee non-nilness
[jsonDict addEntriesFromDictionary: additionDict]
```

geoffmacd: if we are adding nil entries like `[jsonDict addEntriesFromDictionary:nil]`, the json will still be non-nil. so we don't need a `additionDict` to be non-nil

geoffmacd:
@yuxiang-he could I have a review please?

rogebrd (approved):

PR #222, opened 2021-03-11
Fix objc generation for Bytes property and for hash methods with switch statements

YufeiG:
## **Checklist**

**General Contributing**
- Dropboxer

**Is This a Code Change?**
- [x] Code Change

**Validation**
- [ ] Have you ran `tox`?
- [ ] Do the tests pass?

### DIFF
--- a/stone/backends/obj_c_helpers.py
+++ b/stone/backends/obj_c_helpers.py
@@ -32,7 +32,7 @@ from .helpers import split_words
 
 _primitive_table = {
     Boolean: 'NSNumber *',
-    Bytes: 'NSData',
+    Bytes: 'NSString *',
     Float32: 'NSNumber *',
     Float64: 'NSNumber *',
     Int32: 'NSNumber *',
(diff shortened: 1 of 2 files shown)


rogebrd (commented):
Few questions since I am not super familiar with the Obj C Language. 

- Would this be a breaking change? 
- Do we have tests that show what the generated outcome should be? If so can we update those?

rogebrd (approved):

YufeiG:
- These are 2 bug fixes. The current behavior is wrong. Objective-C switch statements require 'break;' between cases or they are fall through by default. And for _bytes_, currently the code generated cannot compile, but because we don't have routes with _bytes_ type this hasn't been an issue. Every time we've added such a property and the objc generator fails, we've changed the route property to _string_. With this change, _bytes_ should generate now as a string.
- There doesn't seem to be any objc or swift stone generator tests. I can look at adding a suite of tests for these languages later. This diff specifically should go in asap because the fall through switch is causing crashes in our app.

PR #282, opened 2022-08-25
Remove use of deprecated mypy option

hauntsaninja:
`show_none_errors` is deprecated, see https://github.com/python/mypy/pull/13507


hauntsaninja:
Don't feel like signing a CLA, but you get the point

greg-db:
Thanks! I'll ask the team to take care of this.

Issue #288, opened 2022-11-21
Python 3.11 support


samschott:
**Describe the bug**
Stone's frontend uses the `inspect.getargspec` method which was deprecated in Python 3.0 and removed in 3.11 (see https://docs.python.org/3.10/library/inspect.html#inspect.getargspec).

https://github.com/dropbox/stone/blob/bdd49bc1017dcea2d193beb1925cded33a276d54/stone/frontend/ir_generator.py#L1077

This results in attribute errors when running stone in Python 3.11.

**Versions**
* What version of the Stone are you using?  3.3.1
* What version of the language are you using? Python 3.11
* What platform are you using? macOS 13

**Additional context**
The recommended migration path is to use `inspect.getfullargspec` as drop-in replacement.

greg-db:
Thanks for the report! I'll ask the team to update that.

fabaff:
For a lot of distributions will `stone` fail to build due to the missing support for Python 3.11. This will lead to broken `dropbox-sdk-python` packages down the road.

greg-db:
@fabaff Thanks for the note!

dvzrv:
Hi! We're currently rebuilding all Python packages against Python 3.11 on Arch Linux. This project is now broken.

Can you please replace the use of `getargspec()` with `getfullargspec()` (see https://docs.python.org/3/whatsnew/3.11.html) and release a new version? Thanks! :)

greg-db:
@dvzrv Thanks for the feedback! This is open with the team. I'll follow up here with any updates.

johnlarkin1:
Hey @samschott @dvzrv - genuinely apologize for the delay here. I did not realize that this was an issue until today (when I came across it myself). I have pushed up PR#299 which will address this. 

I'm working on getting this merged in today! Thank you all for your patience and thanks for pushing up an issue 👍 

https://github.com/dropbox/stone/pull/299

wfraser:
fixed in https://github.com/dropbox/stone/pull/299

PR #299, opened 2023-05-18
Version compatible inspect handling for class parameters

johnlarkin1:
`getargspec` was deprecated in Python 3.0, but finally removed in Python 3.11. For reference: https://github.com/Pylons/pyramid/issues/3688

That seems like a valid amount of time for us to have fixed it.

We can do our best effort and if the user is running a Python version that still has `getargspec` use that, but if they have `getfullargspec` which is available in Python 3.11, we can use that.

## **Checklist**

**General Contributing**

**Is This a Code Change?**
- [x] Code Change

**Validation**
- [ ] Have you ran `tox`?
- [ ] Do the tests pass?

### DIFF
--- a/stone/frontend/ir_generator.py
+++ b/stone/frontend/ir_generator.py
@@ -5,6 +5,7 @@ from collections import defaultdict
 import copy
 import logging
 
+from inspect import isclass
 try:
     from inspect import getfullargspec as get_args
 except ImportError:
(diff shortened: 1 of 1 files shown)


sderickson (approved):
LGTM!

devPalacio:
Fixes #288

PR #307, opened 2023-07-18
Replace uses of deprecated imp with importlib

julianlocke:
Imp is removed in Python3.12, we should use importlib instead. This will also quiet some deprecation warnings.

## **Checklist**

**General Contributing**
- [ ] Have you read the Code of Conduct and signed the [CLA](https://opensource.dropbox.com/cla/)?

**Is This a Code Change?**
- [ ] Code Change

**Validation**
- [ ] Have you ran `tox`?
- [ ] Do the tests pass?

### DIFF
--- a/stone/cli.py
+++ b/stone/cli.py
@@ -168,8 +168,7 @@ def main():
         # The module should should contain an api variable that references a
         # :class:`stone.api.Api` object.
         try:
-            api_module = _load_module('api', args.api[0])
-            api = api_module.api
+            api = _load_module('api', args.api[0]).api
         except ImportError as e:
             print('error: Could not import API description due to:',
                   e, file=sys.stderr)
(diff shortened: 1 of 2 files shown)


stone/cli.py
```diff
@@ -368,6 +367,17 @@ def main():
...
+    file_name = os.path.basename(path)
+    module_name = file_name.replace('.', '_')
+
+    module_specs = importlib.util.spec_from_file_location(module_name, path)
+    module = importlib.util.module_from_spec(module_specs)
+    module_specs.loader.exec_module(module)
+
+    sys.modules[name] = module
+
+    return module
```
devPalacio, line 381: This doesn't work for python <= 3.5, if you want to preserve the ability to run those lower versions, you can borrow inspiration from https://github.com/epfl-scitas/spack/blob/a60ae07083a5744607064221a0cd48204a54394e/lib/spack/llnl/util/lang.py#L598-L625

devPalacio (approved):

devPalacio:
Fixes #295

PR #317, opened 2024-01-04
Generate missing app auth routes, clean up task reconnection helpers

julianlocke:
In the stone intermediate representation of the api, route auth types can either be a string or an array of strings. Previously, stone only identified a route as app-auth if it was exclusively app auth, it never checked to see if the array contained the app auth type. This led to some app auth routes not being included in the app auth clients.

This PR correctly identifies these multi-auth routes as app auth. It also ensures that we don't create duplicate objc compatibility wrappers for routes that are both app auth and some other auth.

It also makes a pair of small changes to background session task reconnection helpers: App auth routes now get their own reconnection helpers, and objc wrappers over reconnection helpers are slightly modified to avoid potentially creating unreachable defaults in switch statements.

See https://github.com/dropbox/SwiftyDropbox/pull/396 for the modified output code.

## **Checklist**

**General Contributing**

**Is This a Code Change?**
- [x] Code Change

**Validation**
- [ ] Have you ran `tox`?
- [x] Do the tests pass?

### DIFF
--- a/stone/backends/swift_client.py
+++ b/stone/backends/swift_client.py
@@ -212,6 +212,7 @@ class SwiftBackend(SwiftBaseBackend):
         template_globals['objc_init_args_to_swift'] = self._objc_init_args_to_swift
         template_globals['objc_result_from_swift'] = self._objc_result_from_swift
         template_globals['objc_no_defualts_func_args'] = self._objc_no_defualts_func_args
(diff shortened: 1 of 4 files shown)


jaquish (approved):
LGTM. I reviewed the corresponding output code

PR #260, opened 2022-01-07
Add sorting of ApiRoute, guaranteeing stable file output.

yanks:
For generations using the JSON-allowlist, the current version does not guarantee that source will be written in the same order. I have addressed this by making sure routes are added in a stable order.

**General Contributing**
- [x] Have you read the Code of Conduct and signed the [CLA](https://opensource.dropbox.com/cla/)?

**Is This a Code Change?**
- [ ] Non-code related change (markdown/git settings etc)
- [x] Code Change
- [ ] Example/Test Code Change

**Validation**
- [x] Have you ran `tox`?
- [x] Do the tests pass?

PR #345, opened 2024-08-23
MF-5759 fix unstable route order for routes with same name but different versions

adecker89:
Looks like there was a previous attempt at this, but it's still unstable
https://github.com/dropbox/stone/pull/260/files

I fixed the compare method added there and removed the sort that was added in favor of updating the existing normalize method to also sort by version

## **Checklist**

**General Contributing**

**Is This a Code Change?**
- [x] Code Change

**Validation**
- [ ] Have you ran `tox`?
- [ ] Do the tests pass?


devPalacio (approved):
This is going to fix so many cache misses!

PR #318, opened 2024-01-24
remove dependency on six

a-detiste:
Your project states "Stone requires Python 3.", so six can be dropped.

https://wiki.debian.org/Python3-six-removal

## **Checklist**

**General Contributing**
- [x] Have you read the Code of Conduct and signed the [CLA](https://opensource.dropbox.com/cla/)?

**Is This a Code Change?**
- [ ] Non-code related change (markdown/git settings etc)
- [x] Code Change
- [ ] Example/Test Code Change

**Validation**
- [x] Have you ran `tox`?
- [?] Do the tests pass?

PR #365, opened 2026-07-07
Modernize for Python 3, drop six dependency

AndreyVMarkelov:
## Summary

Modernizes Stone for Python 3 and removes the `six` compatibility library now that Python 2 is unsupported.

- Replace `six.ensure_str()` with `str()` in the obj_c, swift, and `python_rsrc` backends
- Drop the PY2 utf-8 decode branch in the `String` validator
- Simplify `_strftime()` to `dt.strftime()`, removing the pre-1900 proleptic-Gregorian workaround (and the now-dead `_ILLEGAL_S`/`_findall` helpers and `re`/`time` imports)
- Fix `datetime.utcnow()` deprecation warnings in tests (→ `datetime.now(datetime.UTC)`)
- Set `python_requires='>=3.11'` and refresh trove classifiers (3.11–3.14)
- Bump test deps for modern Python (`coverage>=7.1.0`, `pytest>=7.0.0`) and fix a stray `/coverage` typo in `test/requirements.txt`

Fixes the deprecation-warning report (WEBSERVDB-10027) and overlaps with the intent of #318.

## Testing

All 181 tests pass with **0 warnings** on Python 3.14. Smoke-tested `stone` codegen (python_types + js_types backends) end-to-end: generated code compiles, imports, and round-trips.

## Downstream impact

This release **breaks no existing consumer** — each is insulated from picking it up automatically:

| SDK | How it consumes Stone | `six` removal | `python_requires>=3.11` |
| --- | --- | --- | --- |
| dropbox-sdk-dotnet | PyPI (unpinned), build-time only, generated code committed | N/A (no runtime Python) | Bump `spec_update.yml` Python 3.7 → 3.11 |
| dropbox-sdk-python | PyPI, **runtime** re-export, capped `stone<=3.3.9` | Safe (carries its own `six` dep) | Raise cap + drop EOL 3.9/3.10 CI legs |
| dropbox-sdk-java | git submodule (pinned commit), build-time | Safe (CI installs `six` explicitly) | Bump CI Python 3.9.14 → 3.11 |

Verified none of the three consumers' own generator/runtime code depends on the removed internals. The `>=3.11` floor gates *future adoption* (each repo must modernize its Python to pull the new release), not existing installs — all the pinned Python versions across the fleet (3.7/3.9/3.10) are already EOL.

### DIFF
--- a/stone/backends/obj_c_types.py
+++ b/stone/backends/obj_c_types.py
@@ -1,8 +1,6 @@
 import json
 import os
 
-import six
-
 from stone.backends.obj_c import (
     base_file_comment,
     comment_prefix,
@@ -921,7 +919,7 @@ class ObjCTypesBackend(ObjCBaseBackend):
                          if data_type.min_length else 'nil'),
                         ('maxLength', '@({})'.format(data_type.max_length)
                          if data_type.max_length else 'nil'),
-                        ('pattern', '@"{}"'.format(six.ensure_str(pattern))
+                        ('pattern', '@"{}"'.format(str(pattern))
                          if pattern else 'nil'),
                     ]))
 
--- a/stone/backends/obj_c_types.py
+++ b/stone/backends/obj_c_types.py
@@ -1,6 +1,8 @@
 import json
 import os
 
+from stone.backends.helpers import (
+    ensure_str, )
 from stone.backends.obj_c import (
     base_file_comment,
     comment_prefix,
@@ -54,15 +56,6 @@ if _MYPY:
     import typing  # noqa: F401 # pylint: disable=import-error,unused-import,useless-suppression
 
 import argparse
-
-
-def _ensure_str(value):
-    # Decode bytes to a native (utf-8) str, leaving str values untouched.
-    if isinstance(value, bytes):
-        return value.decode('utf-8')
-    if isinstance(value, str):
-        return value
-    raise TypeError('not expecting type %s' % type(value))
 
 _cmdline_parser = argparse.ArgumentParser(prog='obj-c-types-backend')
 _cmdline_parser.add_argument(
@@ -928,7 +921,7 @@ class ObjCTypesBackend(ObjCBaseBackend):
                          if data_type.min_length else 'nil'),
                         ('maxLength', '@({})'.format(data_type.max_length)
                          if data_type.max_length else 'nil'),
-                        ('pattern', '@"{}"'.format(_ensure_str(pattern))
+                        ('pattern', '@"{}"'.format(ensure_str(pattern))
                          if pattern else 'nil'),
                     ]))
 
--- a/stone/backends/obj_c_types.py
+++ b/stone/backends/obj_c_types.py
@@ -55,6 +55,15 @@ if _MYPY:
 
 import argparse
 
+
+def _ensure_str(value):
+    # Decode bytes to a native (utf-8) str, leaving str values untouched.
+    if isinstance(value, bytes):
+        return value.decode('utf-8')
+    if isinstance(value, str):
+        return value
+    raise TypeError('not expecting type %s' % type(value))
+
 _cmdline_parser = argparse.ArgumentParser(prog='obj-c-types-backend')
 _cmdline_parser.add_argument(
     '-r',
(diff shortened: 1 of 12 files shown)


stone/backends/python_rsrc/stone_serializers.py
```diff
@@ -658,7 +654,7 @@ def decode_union(self, data_type, obj):
         else:
             raise bv.ValidationError("expected string or object, got %s" %
                                      bv.generic_type_name(obj))
-        return data_type.definition(six.ensure_str(tag), val)
+        return data_type.definition(str(tag), val)
```
JukkaL: `str(x)` is not the correct replacement for `six.ensure_str(value)`. The latter means essentially `value.decode("utf-8") if isinstance(value, bytes) else value`.

stone/backends/python_rsrc/stone_serializers.py
```diff
@@ -785,7 +781,7 @@ def decode_union_old(self, data_type, obj):
...
```
JukkaL: Similar to above.

stone/backends/obj_c_types.py
```diff
@@ -921,7 +919,7 @@ def _determine_validator_type(self, data_type, value, has_default):
                          if data_type.min_length else 'nil'),
                         ('maxLength', '@({})'.format(data_type.max_length)
                          if data_type.max_length else 'nil'),
-                        ('pattern', '@"{}"'.format(six.ensure_str(pattern))
+                        ('pattern', '@"{}"'.format(str(pattern))
```
JukkaL: Similar to above.

stone/backends/swift_types.py
```diff
@@ -303,7 +302,7 @@ def _determine_validator_type(self, data_type, value):
                 self._func_args([
                     ("minLength", data_type.min_length),
                     ("maxLength", data_type.max_length),
-                    ("pattern", '"{}"'.format(six.ensure_str(pat)) if pat else None),
+                    ("pattern", '"{}"'.format(str(pat)) if pat else None),
```
JukkaL: Similar to above.

AndreyVMarkelov:
Good catch, thanks @JukkaL — you're right that `str(x)` diverges from `six.ensure_str(x)` on the bytes path (`str(b'abc')` yields `"b'abc'"`). Fixed in 01d9453 by adding an `_ensure_str` helper that decodes bytes as utf-8, passes `str` through, and raises `TypeError` otherwise — matching six's semantics — and applying it to all four call sites (`stone_serializers.py`, `obj_c_types.py`, `swift_types.py`).

Same commit also fixes the CI failures: the `pypy-3.8`/`pypy-3.10` legs choked on `datetime.UTC` (3.11+), so I moved the matrix to 3.11–3.14 to match `python_requires`, and bumped the pinned `flake8` (5.0.4 crashes on 3.12+ due to the removed `ast.Str`).

stone/backends/python_rsrc/stone_serializers.py
```diff
@@ -32,6 +28,15 @@
     import typing  # noqa: F401 # pylint: disable=import-error,unused-import,useless-suppression
 
 
+def _ensure_str(value):
```
JukkaL: Can you used the `ensure_str` function from the helpers module instead?

JukkaL (approved):

stone/backends/python_rsrc/stone_serializers.py
AndreyVMarkelov: Fixed. Thank you!

stone/backends/obj_c_types.py
AndreyVMarkelov: Fixed. Thank you!

stone/backends/swift_types.py
AndreyVMarkelov: Fixed. Thank you!

stone/backends/python_rsrc/stone_serializers.py
AndreyVMarkelov: Fixed. Thank you!

stone/backends/python_rsrc/stone_serializers.py
AndreyVMarkelov: Fixed. Thank you!

AndreyVMarkelov: Done in 76113f9 — now importing `ensure_str` from `stone.backends.helpers`. Confirmed generated SDK code still imports and round-trips, since it already depends on the installed `stone` package via `from stone.backends.python_rsrc... import *`.

cguldner (approved):

